PR #76, opened 2013-03-31
[Design] Implement multi-row-aware left/right cycling through workspaces

autarkper:
If workspace cycling is enabled, and there are multiple workspace rows, the current row is incremented or decremented as appropriate, so that you are never stuck on the same row.

Cycling on up or down is not taken care of (and I don't think it is desirable).


clefebvre:
Closing due to lack of activity/discussion.

PR #156, opened 2014-07-30
Maximize windows that match the work area

jmendeth:
Many applications are buggy: they save the bounds of their window for the next time, but not the maximized state. When you close such an app that is maximized, the next time a window opens that covers the exact size  of the work area, but is not maximized.

An example is Chromium; this is how it opens:

Chromium, unmaximized

In these cases (a maximizable window shows up with a size set to the exact bounds of the work area), the window is automatically maximized. After the change, Chromium shows up as expected:

Chromium, maximized

Please note that, unless the irritating auto-maximize feature that was [removed long time ago](https://github.com/linuxmint/muffin/commit/16d5335ee9d5c4a48e214e98c18f896c1049e09a), this feature is really strict and should go unnoticed for most users. A setting can be created to turn it off, though.

### DIFF
--- a/src/core/window.c
+++ b/src/core/window.c
@@ -3095,6 +3095,18 @@ meta_window_show (MetaWindow *window)
 
   if (!window->placed)
     {
+      if (window->showing_for_first_time && window->has_maximize_func)
+        {
+          MetaRectangle work_area;
+          meta_window_get_work_area_for_monitor (window, window->monitor->number, &work_area);
+          /* Automaximize windows that match the bounds of the work area */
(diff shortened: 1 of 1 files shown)


clefebvre:
mintupdate saves the window size as well, and I can't reproduce the issue with it either.

jmendeth:
Wow, completely forgot about the issue... Sorry guys.

GLolol:
I get this problem with Evince 3.14 and window snapping. Whenever I open a new document, the window opens with a large empty space on the top and left sides, and goes off screen. I think this is because Evince saves its window size, but doesn't remember any window snapping/positioning when it reopens.

PR #170, opened 2015-01-22
Allow programmatic minimization without MWM_FUNC_MINIMIZE

stefand:
This fixes switching out of fullscreen Direct3D applications running in
Wine. See issue 169 for more details.


mtwebster:
Okay got it to work finally off the PPA (1.7.34)

stefand:
Thanks for merging this. Unfortunately I cannot double-check right now because muffin from git fails to compile against the slightly older Cinnamon version Gentoo stable currently ships.

PR #174, opened 2015-02-13
Remove is_ime_popup()

luyikei:
Because both ibus-mozc and fcitx are fixed, this workaround is not needed.

https://code.google.com/p/mozc/issues/detail?id=195
https://github.com/fcitx/fcitx/pull/191

### DIFF
--- a/src/core/window.c
+++ b/src/core/window.c
@@ -8002,19 +8002,6 @@ meta_window_update_struts (MetaWindow *window)
     }
 }
 
-static gboolean
-is_ime_popup (MetaWindow *window)
-{
-    const gchar *icon = window->icon_name;
-    const gchar *wc_name = meta_window_get_wm_class (window);
-    gboolean is_target_name = g_strcmp0 (wc_name, "Main.py") == 0 ||
-                              g_strcmp0 (wc_name, "Ibus-ui-gtk3") == 0;
-
-    gboolean deco = meta_window_get_frame (window) != NULL;
-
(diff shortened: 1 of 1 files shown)


clefebvre:
@luyikei can you confirm which versions of ibus-mozc and fcitx this is fixed since?

luyikei:
Fcitx should be since 4.2.8.6 : http://download.fcitx-im.org/fcitx/
ibus-mozc should be since Mozc OSS Ver. 1.13.1651.102 or r185 : https://code.google.com/p/mozc/source/detail?r=185&path=/trunk/src/renderer/unix/gtk_window_base.cc

mtwebster:
Looks like this is still needed in both Mint 17 (fcitx 4.2.8.3) and LMDE/jessie (4.2.8.5) - might work best to make this something that can be disabled by the user if necessary?

luyikei:
don't necessary, however, it is better to remove unneeded code. This only works for ibus-anthy since ibus-mozc has its own renderer and many users use ibus-mozc or fcitx-mozc. There is no advantages to use ibus-anthy. ibus-mozc, fcitx-mozc or fcitx-anthy are recommended and good choices.

Anyway, since fcitx is not latest version in Linux Mint. This pull request should be postponed, but if fcitx is updated, this pull request can be merged.

clefebvre:
I'll close it for now. LMDE 2 should be a dev. target for another 2 years, and so will Mint 17 until at least 2016.

PR #203, opened 2015-09-01
[Glib connect] prefs: connect to changed:: before reading the value of a setting

leigh123linux:


leigh123linux:
closing as glib2 reverted

PR #218, opened 2016-02-18
Debian improvements

Fantu:
Merged from debian/ubuntu official packages
All these changes should be ok (as already used in debian/ubuntu) and the more probable things to check for this update are:
- if a break/conflict add about latest commit is needed (probably not as is not saw bugreport about muffin migration to ppas in ubuntu)
- if maintscript is needed or /etc/sgml/muffin-common.cat is not present in upstream package


clefebvre:
The following commits were imported into master:

update debian/copyright
debian: fix ppc64el packages build
debian: remove unneeded postinst and prerm

PR #256, opened 2016-08-18
Update _NET_WM_BYPASS_COMPOSITOR implementation to match spec + bug fix

itzexor:
Pull two commits from mutter to update our implementation of _NET_WM_BYPASS_COMPOSITOR to match the spec (thanks @JosephMcc for pointing me to this one) and ensure the value is updated properly when the client only sets the hint during creation.


itzexor:
Errr if you want me to rename that second commit to remove the "x11/" part let me know. It's redundant since we dont support wayland as mutter does.

Issue #268, opened 2016-11-07
Add hide-titlebar-when-maximized Gtk+ hint


lestcape:
A setting for requesting that windows' titlebars should be hidden during maximization has been added to GTK+, please add support for this official Gtk+ configuration property in the muffin window manager, this was adopted by mutter a long time ago:

https://mail.gnome.org/archives/commits-list/2011-December/msg05309.html

Mutter relate code:
https://github.com/GNOME/mutter/search?utf8=%E2%9C%93&q=hide_titlebar_when_maximized&type=Code

PR #269, opened 2016-11-08
[3.4] Support hide-titlebar-when-maximized

lestcape:
Code from gnome shell mutter: https://mail.gnomeorg/archives/commits-list/2011-December/msg05309.html 

This fixed: https://github.com/linuxmint/muffin/issues/268.


JosephMcc:
For reference: https://github.com/linuxmint/muffin/commit/f00ae24efec1bf4457c2b40d854dc46e09260a55

Followed by: https://github.com/linuxmint/muffin/commit/6a6874017ee54279b1c4924fb633fc8aa309892b

lestcape:
jump... :) You are waiting for me? Why this is removed? This is a Gtk+ standard settings...

JosephMcc:
I assume it's an unwanted behavior in Cinnamon. It's up to the window manager to decide weather or not it wants to support this. I think we've had the conversation about this type of thing before. It's not the type of thing we have historically wanted to support in cinnamon.

lestcape:
Well @JosephMcc you assume that or you are sure of that, because is that is corrected, i will remove the global menu from github, as also it`s the same and i don't want to continues in the development of a thing without future that also i don't want, maybe a clone to gnome shell will be better then...

lestcape:
My idea it's after this add a new windows property to be possible get/set this value: https://github.com/lestcape/muffin/blob/a3f06f77b3bc40780f4ed88f6b6bfe4db47e30f3/src/core/window-private.h#L150 from the cinnamon side. So will be possible then modify the value for any windows, not only gtk+ windows also can be removed the atom binding as what i want it's not gtk+ specific, but like this only on windows start don't disturb my idea.

lestcape:
I really think that previous comments of several cinnamons dev's are right, when they say: "we will help to support global menu if some want to do that, but we don't want to add this by ours. This is then only a thing that re-evaluate, what it's the real position about that and be consequent with this... Not more than that...

JosephMcc:
Yeah, I was just chatting with mtwebster about this.

I'm not sure if it would be possible to do what you want ultimately. Just adding this bit won't do it. As I understand this is set at the application level. Adding this part alone I think will leave you with only some windows maximizing with no titlebar which would be strange and inconsistent behavior. I also don't know that this can work with client side decorated applications.

lestcape:
You are right on this, just i don't see that this was removed, if i see this before i will try to add directly what i want and not this... I know that mtwebster don't like this, also i don like this, but wharever the importance of a global menu it's reduce the space, if this is not supported, only the menu portion can be removed and this have not to much sense to me. Thanks.

lestcape:
The idea will work, because it's tested in Arch with Mutter on Gnome, if i'm not sure of that i will not try to add this:
https://wiki.archlinux.org/index.php/GNOME#Hide_titlebar_when_maximized

JosephMcc:
I just built and installed this and currently it doesn't seem to do anything. Not sure it has a lot of purpose currently.

lestcape:
Yes, @JosephMcc, i know that, this occurs because on windows recalc_window_features this value it's not taking in account, because of this:
https://github.com/lestcape/muffin/blob/a3f06f77b3bc40780f4ed88f6b6bfe4db47e30f3/src/core/window.c#L8374

It's just a thing of merged this path with this line https://github.com/jsparber/mutter-hide-legacy-decorations/blob/master/hideTitlebar.patch

lestcape:
I will remove this pull, as i can not be able to decorate a windows properly during a drag and drop. 

On a windows decorated notify event, basically will be replace the meta windows actor by a new one and this break all operation that are grab from the display. The solution of only ensure a frame where there are a not grab operation is possible, but this not a good implementation or something that i want to have (an undecorated windows in the desktop during a drag and drop).

If will be finished the grab operation, the windows will be dropped and also this is not good.

To implemented this behavior in a consistent way, will need to be added a lot of code that probably will introduced more bugs.

As i have not an easy way to implement this as it's in mutter, probably because some required code was removed or because the branch where muffin come are old and this won't work ok. I don't know...

PR #272, opened 2016-11-14
[3.4] Support hide the windows titlebar

lestcape:
This pull add the ability to hide a muffin window titlebar.

It's possible hide an individual window titlebar in any moment and also it's possible alert muffin to hide the title bar of a windows, when the windows will be maximized.  Both actions do nothing is the windows it's client-side decorated.

The idea it's remove the windows border instead of undecorate the windows, as undecorate a windows have a side effect, please see:
https://github.com/linuxmint/muffin/pull/269

This pull add the support for an upcoming pull request on cinnamon side that add the ability for hide the titlebar of all windows when they are maximized:
https://github.com/linuxmint/Cinnamon/pull/5961

So enable hide the titlebar of a windows when maximized, will be easy as do in your cjs code:
```
Main.wm.hideTitlebarWhenMaximized(true);
```
And for revert this:
```
Main.wm.hideTitlebarWhenMaximized(false);
```


src/core/window.c
```diff
@@ -8345,6 +8352,65 @@ set_allowed_actions_hint (MetaWindow *window)
...
+ * Return value: %TRUE if the window titlebar is hidde.
+ */
+gboolean
+meta_window_is_titlebar_hide (MetaWindow *window) {
+   if (!meta_window_is_client_decorated (window))
+        return ((window->hide_titlebar_when_maximized &&
+           META_WINDOW_MAXIMIZED (window)) ||
+           window->hide_titlebar);
+  return FALSE;
+}
```
JosephMcc: Something looks wrong with indentation here. It should be using an indentation of 2. I would also left align the three values in you return statement.

lestcape:
Thanks @JosephMcc the code follow your recommendation now.

PR #294, opened 2017-09-06
Improve debian packaging

Fantu:


clefebvre:
I pushed "Remove old things from initial fork that seems not needed" to master.

I don't really see the point of the other commits, they don't seem to bring anything useful to us.

Fantu I find it hard to merge your PRs because you mix different commits into them. Debian isn't upstream to our packaging, it's the other way around and we're not interested in merging our packaging blindly like this without justifications just to make it the same as what happened elsewhere.

Please separate your commits into separate PRs going forward and justify changes, explaining why they're beneficial outside of the scope of Debian.

Fantu:
Is not easy and fast separate in minimal splitted PR with many changes and I not have time to test all cases, I created PR with full (or near in some cases) debian things already tested.
Not related to only this but all packages there are some things important that can solves some effettive problem or help to avoid it, there are also many other things out of debian packaging important but not considered by "mint" software until the problem happen, things that can be easy prevented considering deprecated thing, removing/replacing them ASAP.
About other things of debian packaging I thing can be useful keep a look to debian packing done also by expert debian developer like maxy and marga, tested on newer versions (with debian unstable) that will be used, trying to make packaging better and solid (doing it preferably during development)
Similar thing with upstream software forked from, I saw that is already done when some problem happen and some other cases but unfortunatly now seems not be enough to be near level of main DE :(

These problem seems one of the mainly things that cause me to avoid install linux mint to dozens of workstations, before initial major packaging improvement of some years ago also on desktop with minimal usage because mint had packages upgrade problem in many cases that I had to solves manually (recently I'm retrying on few installation) and mint can't be chosen in installations I do with linux mainly for have a system better that windows and with almost no problem in comparison, I use instead mainly ubuntu and other derivates with more initial changes (to have major of needed things "out of the box") but less risk of problem that help me to have dozens of users happier and decreasing the average of 5-10 yearly technical interventions mainly software related to max 1-2 per year mainly hardware related.

Unfortunately I not have enough time and knowledge for give significative help to make mint the better choice (good and reliable) out of the box at least for noob users with minimal usage.

clefebvre:
Wow.. seriously? You come up with a bunch of changes, all mixed up in one single PR, when asked to explain them and make things modular you get frustrated and start criticizing Mint? What's next Fantu, you're going to go personal and criticize me personally next? Wtf seriously. We're supposed to just copy-paste whatever changes you made in Debian, turn a blind eye to what looks peculiar to us and you don't get to explain anything?

We take upgrades very seriously. If we didn't we would merge your PRs without caring what's in them.

I've personally fixed upgrade issues caused by your PRs in the past. That doesn't mean you're a poor developer, and I don't need to go on a leap and criticize you and Debian out of frustration, just like you did there. No. Everybody does it, everybody breaks stuff now and then.

You don't get a free pass because this "is coming from Debian". Keep packaging simple, avoid making unnecessary changes, make small/modular PRs and always explain your changes.

Adhere to these development principles and you will be taken more seriously.

Fantu:
@clefebvre Sorry for my bad english and for thing not write good or even the opposite of what I have in mind.
I know that I'm not good knowledge about some things and not enough time to give significant and good help. Also explain in details most of things in english seems a problem for me :(
I not critical with this and I not want force to accept my PR, significant rebase can require more check but if I can I do them thinking to be useful.
Linux Mint is a good project even if some occasional important issue more than ubuntu and the idea of good distro out of box for all main basic things easy also for noobs is optiomal and really need.
Cinnamon also is good, I use it on my workstation for some years, after gnome2 I not found better DE than cinnamon for working.
I'm sorry which lacks sufficient help to keep track of upstream softwares forked for easy take all useful bugfix, backport useful/needed technologies ASAP before having significant issue/lack instead after.
Packaging also have some "lack" but is minor has less global impact except for rare cases.
PS: I also apologize for the "insult" that made assume that Mint is not at least better than Windows as opposed to other gnu/linux distributions ^^'' In reality I think Mint is one of the few derivates that to have a sense and to be good.
I hope Mint will reach the solidity of debian stable, with at least full common basic usage (and more) out of the box of what is needed and expect average users who are used only Windows (probably is already more of what I suppose)

PR #297, opened 2017-09-20
Reduce window shadow opacity

krogank9:
Window shadows for selected windows are too opaque, they don't match most default themes. For instance, the title bar text on the default theme changes from #b2b2b2 -> #999999, a 10% difference. The window shadows however go from 50% opacity to 100% on selected windows. This change in opacity is especially noticeable when monitor brightness is set to low and should be reduced to make it easier on the eyes. This changes window shadows from 50% opacity on unselected to 80% opacity when selected, making them easier on the eyes without sacrificing any usability or clarity of which window is selected.

Before: https://i.imgur.com/YhqqCO8.png
After: https://i.imgur.com/01OXsLI.png

### DIFF
--- a/src/compositor/meta-shadow-factory.c
+++ b/src/compositor/meta-shadow-factory.c
@@ -127,17 +127,17 @@ static guint signals[LAST_SIGNAL] = { 0 };
 /* The first element in this array also defines the default parameters
  * for newly created classes */
 static MetaShadowClassInfo default_shadow_classes[] = {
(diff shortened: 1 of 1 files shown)


JosephMcc:
Have to admit I'm not a huge fan of this change. If we were looking to change the shadow style I'd prefer to something more like the Mac with larger more blurred shadows that gave the appearance of more "depth".

Issue #302, opened 2017-11-21
Window animations on 2d grid are in wrong direction


gunsch:
```
 * muffin version (muffin --version or 'dpkg --list | grep libmuffin0' for Mint/Ubuntu)
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...)
 * Graphics hardware *and* driver used
 * 32 or 64 bit
 ```

muffin version: muffin 3.2.1
Distribution: Debian GNU/Linux rodete (Google internal build)
`lspci | grep -i vga`: VGA compatible controller: NVIDIA Corporation GM107GL [Quadro K620] (rev a2)
64-bit

**Issue**
Window animation moves in incorrect direction when switching workspaces in a 2d grid.

Prerequisites:
* Window Effects are enabled
* Using the Cinnamon applet [`workspace-grid@hernejj`](https://github.com/hernejj/workspace-grid-cinnamon-applet/tree/master/workspace-grid%40hernejj) (seems to be the de-facto applet for this functionality)

**Steps to reproduce**

* Set up workspaces as 2d grid with >1 row of workspaces. (I use 3x3).
* Use the "Ctrl+alt+down" and "Ctrl+alt+up" 
* Ctrl+alt+down animation moves "down and to the left", and Ctrl+alt+up animation moves "up and to the left".

**Expected behaviour**
Down/up animations move directly down and up.

**Other information**
Others have reported this bug publicly, but not in muffin, e.g. on the [workspace applet](https://github.com/hernejj/workspace-grid-cinnamon-applet/issues/19), but from some initial investigation I believe the problem is in muffin's code.

1. The applet (Cinnamon JS applet) simply calls `Main.wm.actionMoveWorkspaceDown` and `Main.wm.actionMoveWorkspaceUp`, where `Main` is `imports.ui.main`.
1. `main.js` has the following definitions:
    ```
    actionMoveWorkspaceUp: function() {
        global.screen.get_active_workspace().get_neighbor(Meta.MotionDirection.UP).activate(global.get_current_time());
    },
    actionMoveWorkspaceDown: function() {
        global.screen.get_active_workspace().get_neighbor(Meta.MotionDirection.DOWN).activate(global.get_current_time());
    },```
[...]

PR #303, opened 2017-11-21
Fixes workspace animation direction for grid systems.

gunsch:
Right now, up/down animations proceed at angled directions. This seems
to be because |get_wrapped_horizontal_direction| assumes that |from|
and |to| are always different values.

In theory, this should resolve issue #302. However, I'm having a hard time figuring out how to verify this in a full cinnamon runtime. Can I run my locally-built muffin replacement with my stable cinnamon build? Do I need to do a full cinnamon build + all of its dependencies to be able to verify this muffin change? Some pointers would be appreciated. Thanks!

### DIFF
--- a/src/core/workspace.c
+++ b/src/core/workspace.c
@@ -547,7 +547,7 @@ get_wrapped_horizontal_direction (gint                from,
           ret = (to - from) <= ((num_workspaces - to) + from) ? META_MOTION_LEFT : META_MOTION_RIGHT;
         else
           ret = META_MOTION_LEFT;
-      else
+      else if (from > to)
         if (wrap)
           ret = (from - to) <= ((num_workspaces - from) + to) ? META_MOTION_RIGHT : META_MOTION_LEFT;
         else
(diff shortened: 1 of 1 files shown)


JosephMcc:
All you need to do is rebuild and install muffin with your changes. Then restart Cinnamon.

gunsch:
Confirmed my fix works as expected! Restarting cinnamon in-place had some difficulties but I was able to verify it after replacing libmuffin.so + rebooting.

gunsch:
Hey, ping on this --- anything I can do to make this easier to review? I've been running my compiled libmuffin.so on my desktop for the last two weeks with this fix and it's worked great.

collinss:
@gunsch we're just finishing up a dev cycle. It might be a little bit still before stuff get merged for the next release. I have a number of pr's open in Cinnamon myself.

gunsch:
One more ping on this, since it's been awhile :)

Tito1337:
Thank you @gunsch, I'm running Cinnamon with your patch and it's working great.

PR #319, opened 2018-05-06
Use lighter shadows

germanfr:
See the reasoning here: https://bugzilla.gnome.org/show_bug.cgi?id=702141

**>> Before**
Before
**>> After**
After

### DIFF
--- a/src/compositor/meta-shadow-factory.c
+++ b/src/compositor/meta-shadow-factory.c
@@ -127,17 +127,17 @@ static guint signals[LAST_SIGNAL] = { 0 };
 /* The first element in this array also defines the default parameters
  * for newly created classes */
 static MetaShadowClassInfo default_shadow_classes[] = {
-  { "normal",       { 6, -1, 0, 3, 128 }, { 3, -1, 0, 2, 64 } },
-  { "dialog",       { 6, -1, 0, 3, 128 }, { 3, -1, 0, 2, 64 } },
-  { "modal_dialog", { 6, -1, 0, 3, 128 }, { 3, -1, 0, 2, 64 } },
-  { "utility",      { 3, -1, 0, 3, 128 }, { 3, -1, 0, 2, 64 } },
-  { "border",       { 6, -1, 0, 3, 128 }, { 3, -1, 0, 2, 64 } },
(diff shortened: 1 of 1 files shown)


JosephMcc:
I really don't like this idea. If we were going to change the shadows, I'd much prefer to see us go with wider, more blurred shadows. The MacOS does. It makes the active window far more distinguishable.

JosephMcc:
Most Gtk themes also seem to use large, more blurred shadows on csd windows and this really makes them match less.

JosephMcc:
Your link doesn't seem to exist ;)

germanfr:
Oh, sorry. It was mutter not gnome-shell. This is it: GNOME/mutter@91c6a144da53aab49816f4781dfab6e91e248358

JosephMcc:
So I like the type of shadow in that second one much better. It's still just a little too light for my taste. Unfocused windows also have essentially nothing.

germanfr:
Better now? I've increased size and made them little bit darker :)

JosephMcc:
Cool, I'll try it again tonight. I'm not against the idea. I've even played with it in the past. It's just one of those things that I wouldn't want to see merged right now, but instead done during the next dev cycle. I'd like to actually use it that way for awhile and see what I think. I'm also sure others are going to have their own opinions on the subject.

JosephMcc:
I think that's better but here are a couple of screenshots of what I was talking about. See how there is a really nice depth to it? Even the unfocused windows have some, but the focused window really stands out.

macos-fat

yosemite-980x612

clefebvre:
Hi guys,

Is this still relevant once we move towards GTK windows?

If not we don't need to argue at all over this right?

germanfr:
I guess not. Gnome removed this part of the code from mutter long ago for a similar reason, but idk for sure.

JosephMcc:
It would still be relevant since muffin would continue to draw shadows for sever side decorations. The drawing would just be done with info pulled from the gtk theme instead of a separate metacity theme.

It's really just a matter of personal taste. Most themes tend to have larger blurred shadows on client side decorated windows vs. the ones drawn by the window manager. At based on my own observations. Lightening them and making them smaller would just make it more apparent, imho.

clefebvre:
We'll tackle this in 4.2, along with GTK theming.

PR #324, opened 2018-06-02
[Experimental] theme: Draw window borders with the gtk3 theme

JosephMcc:
This drops all support for the metacity theme format. Borders will now drawn
by the gtk theme. This is a rough implementation still and some decisions need
to be made with regards to handling things like fringe button support.

Thanks to the upstream mutter devs who did all the original work this is
based on.


JosephMcc:
Replace by: https://github.com/linuxmint/muffin/pull/371

PR #330, opened 2018-07-08
[WIP] Meson port

Salamandar:
Well, you know what it's about.

For this port I also deleted the Autotools files. You seem to delete them just after my PR in the other repos…


Salamandar:
Well I forgot to update the debian/rules too.

Salamandar:
CI fails because of required meson_version `0.46.0`… I guess muffin-meson will not be for Mint 19.

leigh123linux:
@Salamandar 

This PR will probably sit here for 5 years unless you relax the meson requirement.

> CI fails because of required meson_version 0.46.0… I guess muffin-meson will not be for Mint 19

Salamandar:
What minimum version would you expect for Mint19 ? Mint20 ? 
I could get the required version down to `0.41.0` with an ugly (but working) hack around the pkgconfig generation.

clefebvre:
0.45 in Mint, but 0.37 in LMDE, we need to support both.

clefebvre:
@Salamandar we're at 0.47 now in LMDE, so 0.45 is the min version we care about.

ItzSwirlz:
This PR is getting old-I suggest when Mint 21 gets released and whenever all Cinnamon/muffin/cinnamon bundle packages stop getting 4.8 point releases merge this and make sure it works.

I also then want to ensure we have a meson port for the X-Apps and any other missing things.

For python packages like hypnotix/warpinator, it **may** make sense to add a setup.py or a meson build to generate a ninja file to install the package locally which does have debian packaging integration.

Salamandar:
Yeah, I'm working on getting this PR up to date.

PR #347, opened 2018-09-11
[WIP] Cache the composite overlay window

jaszhix:
XCompositeGetOverlayWindow was getting called on every window mapped.

Partially based on:
https://github.com/GNOME/mutter/commit/05899596d10918df5359d89baa82e6fedd0ae208


jaszhix:
This PR causes applications to segfault, so abandoning for now.

PR #350, opened 2018-09-13
Fix desktop background appearing twice on initial start up

jaszhix:
This corrects a regression from https://github.com/linuxmint/muffin/commit/ca08977fb5b62b22683f40b27633adcd8812aed5.


clefebvre:
BG still appears. We need to call redirect_windows() first thing.

PR #351, opened 2018-09-14
[WIP] Use utility functions to convert between frame and client rectangles

jaszhix:
Based on:
https://github.com/GNOME/mutter/commit/f36a62733098d8ef416d627c20f50c64bc7f0a4b

It appears *not* implementing the changes in constraints.c has a performance benefit, and reduces input latency between the corner of a window and the cursor while resizing windows - compared to with the patch fully implemented, and how it was before.

We're passing cached borders to `meta_window_unextend_by_frame` and `meta_window_extend_by_frame` in constraints.c. The drop-in replacements for those functions recalculates the window borders for windows with frames, but it doesn't seem to be necessary to do it every time since `meta_window_constrain` is called after `meta_frame_calc_borders` any time a window is moved or resized.

Not seeing any regressions, including quarter tiling and snapping.

### DIFF
--- a/src/core/display.c
+++ b/src/core/display.c
@@ -1539,6 +1539,7 @@ window_raise_with_delay_callback (void *data)
     {
       int x, y, root_x, root_y;
       Window root, child;
+      MetaRectangle frame_rect;
       unsigned int mask;
       gboolean same_screen;
       gboolean point_in_window;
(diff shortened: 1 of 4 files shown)


jaszhix:
Might redo this one later, probably in an effort to port changes from #324.

PR #374, opened 2018-11-04
always process clutter Makefile.in

eli-schwartz:
The current check includes the subdir unless it's explicitly disabled, but only builds the Makefile if it's explicitly disabled. This means the entire project fails to build while processing the docs folder, even if docs are disabled and a no-op... unless clutter docs are separately disabled.


eli-schwartz:
@jaszhix autotools sucks, because you need to process the file even if you have no intention to use it -- the docs directory with --disable-gtk-doc just does "nothing to do for target all", but it still needs to have a Makefile to ignore.

Presumably once this too is ported to meson, we will have some sanity. :)

PR #387, opened 2018-11-22
Revert "clutter: Deliver events sooner when possible"

jaszhix:
This reverts commit 2e88f14d3ac385f3448fd96c5db102cf3e7f28ce.

This patch causes a [regression](https://gitlab.gnome.org/GNOME/mutter/merge_requests/168) with dragging operations when mice run at a 1000hz polling rate.


leigh123linux:
+1 for removing this troublesome mutter commit.

PR #403, opened 2019-01-01
[Next] Reduce usage of signals between internal components

jaszhix:
This is an extension of the idea from [Cinnamon PR #8230](https://github.com/linuxmint/Cinnamon/pull/8230). This uses a callback for MetaWindow's `position-changed` signal, which is a new addition since 4.0, and all it does is notify MetaWindowActor to emit its signal. The signal is replaced, but not removed so it doesn't conflict with #399.

I also have tested replacing `after-paint` and `presented` on ClutterStage, and that didn't work so well because it relies on the deferred behavior from `g_signal_connect_after`.


jaszhix:
Included in #410.

PR #407, opened 2019-01-05
[Next] Avoid usage of g_getenv in frequently called functions

jaszhix:
After looking at the complexity of the function in GLib, it looks like this shouldn't be called inside paint functions.


jaszhix:
Included in #410.

PR #408, opened 2019-01-05
Window-actor: Only apply desaturation effect if the window has a shadow 

jaszhix:
The effect was added to soften shadows in 92b32be. This makes it so it's only applied to windows with shadows.

Depends on #405 to avoid a conflict.


jaszhix:
Included in #410.

PR #410, opened 2019-01-07
Various compositor improvements

jaszhix:
- Merges MetaShapedTexture into MetaWindowActor

This is being done because MetaShapedTexture is painting entirely with Cogl and Cairo - there's nothing requiring it to be a child of MetaWindowActor, or to be a Clutter specific instance. Clutter is just the mechanism used to integrate the compositor into the UI layer, and structure its life cycle methods. The more actors involved in rendering a single window, the more things that can go wrong - like windows flickering from broken actor synchronization - which I haven't seen occur on this branch.

Having all compositing handled on one actor allows reducing functions that are partially doing the same things and reducing function getters (e.g., we now have immediate access to textures).

- Optimizes rendering for completely obscured windows. Currently, the visible property only stays in sync with whether or not the actor is visible, but doesn't take into account if the window is unminimized but completely obscured by another window on top. This reduces the process time spent on these windows.

  - Checks if the actor is obscured when the stack is synchronized and caches the value.
  - Various paint functions will return early if a window is obscured.
  - In order to not break live thumbnail previews, a method is exposed to GI that will override the optimization, but it is the caller's responsibility to make sure that it toggles the override off (obscured == TRUE). The previous obscured state will be cached and restored if true is passed to `meta_window_actor_set_obscured`. All widgets using live thumbnails will need to call this function when showing a window texture, so obscured windows are not treated like minimized ones (frozen texture).

Depends on https://github.com/linuxmint/Cinnamon/pull/8269, https://github.com/linuxmint/Cinnamon/pull/8300

Thanks to @itzexor for helping solve issues with mipmapping and the paint volume.


jaszhix:
Rebased and squashed a few commits together that were just fixing up previous commits in this PR.

clefebvre:
- It feels much snappier :slightly_smiling_face:
- Moving workspaces back and forth, all my windows are missing (as if they were all minimized) when I come back
- The tiling OSD animation looks wrong, it's starting from the center of my left monitor, instead of starting from the edge.... say I tile on the right edge of my right monitor, the animation starts all the way from the center of my left monitor
- If I focus a window and invoke scale (ctrl-alt-down), scale looks good, if I focus a window, move it a bit, then invoke scale, that window isn't fitting within the selection border in scale
- If a video is minimized, its thumbs (in alt-tab, scale etc..) don't animate (that's probably by design?) (edited)

jaszhix:
> Moving workspaces back and forth, all my windows are missing (as if they were all minimized) when I come back

This is a bug I spent time trying to fix, @brownsr identified it as well, and up until now haven't been reproducing it. I'll try to again.

> The tiling OSD animation looks wrong, it's starting from the center of my left monitor, instead of starting from the edge.... say I tile on the right edge of my right monitor, the animation starts all the way from the center of my left monitor

I'll take a look at what values Cinnamon is using to compose this animation.

> If I focus a window and invoke scale (ctrl-alt-down), scale looks good, if I focus a window, move it a bit, then invoke scale, that window isn't fitting within the selection border in scale

https://github.com/linuxmint/Cinnamon/pull/8397

> If a video is minimized, its thumbs (in alt-tab, scale etc..) don't animate (that's probably by design?) 

Yeah, I think that is the case on master as well, and it has to do with the window being frozen - which is a separate mechanism in MetaWindowActor that I tried to not affect because it's tied into MetaSyncRing.

clefebvre:
> The tiling OSD animation looks wrong, it's starting from the center of my left monitor, instead of starting from the edge.... say I tile on the right edge of my right monitor, the animation starts all the way from the center of my left monitor

It looks like this isn't specific to OSD. I can reproduce it with windows as well now. Here's how I do:

- Two monitors set up
- Create a window on the left monitor
- Move it to the right monitor
- Press Windows+Right to tile it to the right

It tiles on the right edge of the left monitor, instead of the right edge of the right monitor. You can reproduce it both ways.. basically, it doesn't know it moved monitor, it still thinks it's on the original one.

jaszhix:
I don't think that issue is caused by this PR.

https://github.com/linuxmint/Cinnamon/issues/4526

jaszhix:
Ok, I'm eating my words, the tiling bug is related to #427. b019ef5dd8c0d61042d27117ad71f81c105ed501 ensures we're basing the monitor calculation on a fresh outer rect value, and moves the `meta_window_update_monitor` call in `update_move`  inside the condition where the rect would actually change. That's needed so this doesn't regress performance, we don't want to be calling `meta_window_update_monitor` unless the window actually moved anyway.

clefebvre:
>     Moving workspaces back and forth, all my windows are missing (as if they were all minimized) when I come back

Fixed

>     The tiling OSD animation looks wrong, it's starting from the center of my left monitor, instead of starting from the edge.... say I tile on the right edge of my right monitor, the animation starts all the way from the center of my left monitor

Also on master.

>     If I focus a window and invoke scale (ctrl-alt-down), scale looks good, if I focus a window, move it a bit, then invoke scale, that window isn't fitting within the selection border in scale

Fixed

> If a video is minimized, its thumbs (in alt-tab, scale etc..) don't animate (that's probably by design?) 

Not really an issue and also on master

Issue #398, opened 2018-12-16
Moving windows make them disappear/blink


jaszhix:
Moved from linuxmint/Cinnamon#8052, @bitstrings wrote: 
#
```
 * Cinnamon version 4.0.2
 * Distribution - Manjaro
 * AMD 580 GPU (amdgpu)
 * 64 bit
 * kernel 4.19.2
 ```

Moving windows make them disappear.

Video (4K Screen) -> [HERE](https://drive.google.com/file/d/1G_WCjCnWB7JH8Ofo2Gj6CcB831jYUTEB/view?usp=sharing)

The is NO CLUTTER_* in environment.

In fact this is a fresh install without mod.

I've got a AMD 580 GPU using the free amdgpu.

Also, performance is iffy. Things are a bit jerky. Tried everything.

With or without xorg.conf.

xorg.conf
```
Section "Device"
	Identifier  "Card1"
	Driver      "amdgpu"
	VendorName  "AMD"
	BoardName   "Radeon RX 580"

    Option "TearFree" "on"
    Option "DRI" "3"
    Option "EnablePageFlip" "on"
EndSection
```

rsi9x:
I am having similar rendering issues after upgrading to Mint Cinnamon from 19 to 19.1.  In my case, the problem seems to affect only the second and third monitors.  It seems that the left-most monitor is unaffected (regardless if it's the primary monitor or not).  Whichever monitor I set as the left-most monitor is always unaffected.  On the center/right monitors, when drag a window to it, sometimes the window (or parts of the window) will flash or completely disappear.  Other times, the decoration remains, but the contents disappear (see video

[mint191_glitching_window_960.mp4.zip](https://github.com/linuxmint/muffin/files/2768218/mint191_glitching_window_960.mp4.zip)

).  
When I interact with the window, the contents usually reappear, but may disappear again if I drag it elsewhere. The desktop wallpaper on the affected monitors will sometimes also randomly flicker. The problem seems to go away as long as a window remains maximized on the affected monitors.

Things I've tried to no avail:
- Removing desktop background
- Changing display resolution
- Reducing number of monitors to 2
- Changing which monitor was primary 
- Both Open Source and Proprietary Nvidia drivers (390 ~ 415)
- Toggled "Enable VBlank" in General settings
- Enabled full composition pipeline in Nvidia settings
- Fresh install of Mint 19.1
[...]

PR #415, opened 2019-01-19
window-actor: Only fall back to shadows if the window has a shape

jaszhix:
This includes all of the windows I think we would want to have shadows, such as framed windows, GTK CSD windows, and chromium windows. 

This removes shadows on windows such as Steam, which has artifacts because it initially maps its menus outside of the parent window causing it to lose focus, which makes the shadow border flicker between focused and unfocused while hovering over selections.

Ref #398

### DIFF
--- a/src/compositor/meta-window-actor.c
+++ b/src/compositor/meta-window-actor.c
@@ -960,7 +960,7 @@ meta_window_actor_has_shadow (MetaWindowActor *self)
     return TRUE;
 #endif
 
-  return TRUE;
+  return priv->window->has_shape;
 }
 
 /**


clefebvre:
I see two issues with Steam and they're both visible in the video.

https://github.com/linuxmint/muffin/files/2768218/mint191_glitching_window_960.mp4.zip

- The most visible issue is that both Steam itself and its menus act like windows with shadows, so when you open a menu in Steam, muffin thinks the main Steam window isn't focused anymore and draws a smaller shadow than before... this is fixed by this PR as it removes the shadow on the main window.

- Artifacts show in the top left corner of the screen. This is only partially fixed by this PR. Although the main window loses its shadows, some of the menus still generate shadows.

jaszhix:
I think this could be fixed by doing a full revert of 9ce812547b30aa5af8e7cc173aea6a2c66f37a8d.

I can't spend any more energy on this, especially when I think shadows look stupid on CSD windows that are not intended to have any.

clefebvre:
OK, just tested a revert of https://github.com/linuxmint/muffin/commit/9ce812547b30aa5af8e7cc173aea6a2c66f37a8d, it affects chrome/chromium but has no impact on steam and doesn't fix this bug.

I also checked steam under Marco and Metacity, both render shadows not only for Steam, but for its menus... and they both have the same issue as Muffin, i.e. they both suffer from https://github.com/linuxmint/muffin/files/2771746/mint191_menu_glitch.mp4.zip

clefebvre:
I'm going to close this. I think since it affects all DEs, it's up to Steam to fix it. They should use proper menus really, not defocus their main windows like they're doing now.

jaszhix:
I think it's still a Muffin issue. Steam does not create shadows for its menus. which means we're applying shadows to window types we shouldn't be. This is not a Steam bug because as far as they're concerned those window menus are not visible in other WMs.

jaszhix:
This PR should remain open because it is fixing an entirely different issue than the shadows appearing in the corner, and that issue could be added to this PR. This fixes the shadow flicker from unfocus. Again, this has nothing to do with Steam.

jaszhix:
https://github.com/ValveSoftware/steam-for-linux/issues/6

PR #417, opened 2019-01-22
meta-shadow-factory: change a routine to use unsigneds

brownsr:
it's clearly intended to take positive values, as it will infinite loop on negatives

### DIFF
--- a/src/compositor/meta-shadow-factory.c
+++ b/src/compositor/meta-shadow-factory.c
@@ -635,8 +635,8 @@ fade_bytes (guchar *bytes,
  */
 static guchar *
 flip_buffer (guchar *buffer,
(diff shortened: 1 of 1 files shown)


src/compositor/meta-shadow-factory.c
```diff
@@ -743,6 +743,9 @@ make_shadow (MetaShadow     *shadow,
   if (buffer_width < buffer_height && buffer_width > (3 * buffer_height) / 4)
     buffer_width = buffer_height;
 
+  g_warn_if_fail(buffer_width >= 0);
```
jaszhix: Do we need to warn about this condition? `make_shadow` will get called before every paint.

PR #427, opened 2019-02-16
window: Cache the outer, client rects

jaszhix:
Split from #410.

This caches the outer and client rects, and updates them when the window size changes. The placement of the update calls were chosen based on a [previous upstream optimization](https://github.com/linuxmint/muffin/commit/8779c27f78150970b98e5681acfb5b792d8fea35) for caching frame borders.

### DIFF
--- a/src/compositor/compositor.c
+++ b/src/compositor/compositor.c
@@ -724,14 +724,11 @@ meta_shape_cow_for_window (MetaScreen *screen,
       XserverRegion output_region;
       XRectangle screen_rect, window_bounds;
       int width, height;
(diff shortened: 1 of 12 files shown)


clefebvre:
Regression when resizing Steam to make it smaller. The space previously used by Steam shows the desktop bg:

image 5

jaszhix:
Pushed https://github.com/linuxmint/muffin/pull/427/commits/7a20d32f9bbe6779c89c614d33d07f21863bdd3c which fixes the Steam artifacts.

PR #429, opened 2019-02-17
Avoid getting context from Clutter in frequently called functions

jaszhix:
Split from #410.

This reduces calls to `clutter_get_default_backend`, which calls `_clutter_context_get_default`, and causes a synchronous lock to be placed on ClutterContext - this isn't necessary for Muffin's configuration.

Initially tried to repurpose the static pointer in cogl-utils.c, but this resulted in a massive memory leak. This adds a getter to MetaCompositor and uses its reference instead.

Also makes hardware_supports_npot_sizes public so it can only be called once when GL extensions are checked.

### DIFF
--- a/src/compositor/cogl-utils.c
+++ b/src/compositor/cogl-utils.c
@@ -26,6 +26,7 @@
 #endif
 #include <clutter/clutter.h>
 #include "cogl-utils.h"
+#include "meta-sync-ring.h"
 #include <meta/errors.h>
 
 #include <gdk/gdk.h>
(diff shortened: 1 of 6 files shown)


clefebvre:
Here's the lock in Clutter: https://github.com/linuxmint/muffin/blob/master/clutter/clutter/clutter-main.c#L1246

jaszhix:
Here are some benchmarks using gputest with Phoronix Test Suite. This measures master (cinnamon and muffin) with this PR and #427 (with cinnamon master). These tests don't accurately measure latency, but the compositor's impact on application draw performance. I think these are accurate enough for what they're measuring because they require no automated interaction to run.

https://openbenchmarking.org/result/1902215-SP-CACHEOUTE71

PR #436, opened 2019-03-14
[WIP] window-actor: Set the actor size once during the map effect

jaszhix:
This fixes texture artifacts that occur when mintupdate (and potentially other apps) maps with effects enabled, but the animation is still broken. It looks like the window is being mapped off-screen and the animation has a greater distance to travel.


jaszhix:
Better solution found in #437, which is always basing the actor size on the texture size.

PR #437, opened 2019-03-16
[WIP] window-actor: Fix and refactor window effects handling

jaszhix:


JosephMcc:
Running this locally. It's an improvement over master. Mintupdate is better but still initially maps to the wrong position before moving to the correct place. But the weird texture/shadow issue is fixed in this branch.

The window close animations are also improved over master but still seem to be lacking some smoothness we had a higher speeds in 4.0

clefebvre:
Quick note about mintupdate, it has code to move/resize its own window... so it's possible it looked like that in 4.0 too.

clefebvre:
@jaszhix same observation for me than @JosephMcc in regards to mintupdate (and pkexec xed), the window/dialog isn't squashed anymore, but the window moves after it's mapped.

This doesn't happen with the 4.0 build.

jaszhix:
Replaced by #449, #453, #460, #463, and #464.

PR #449, opened 2019-03-28
window-actor: override Clutter opacity property and method

jaszhix:
This restores actor opacity management without relying on a signal.

- Merges `maybe_desaturate_window` into `meta_window_actor_set_opacity`.
- Only propagates to clutter if the opacity changes.
- Scrolling on title bars opacifies windows unlike the version of this patch from #437.

### DIFF
--- a/src/compositor/meta-window-actor.c
+++ b/src/compositor/meta-window-actor.c
@@ -53,6 +53,8 @@
  */
 #define MIN_FAST_UPDATES_BEFORE_UNMIPMAP 20
 
+#define META_WINDOW_ACTOR_PARAM_ANIMATABLE (1 << G_PARAM_USER_SHIFT)
+
 enum {
   POSITION_CHANGED,
   SIZE_CHANGED,
@@ -210,7 +212,8 @@ enum
   PROP_X_WINDOW_ATTRIBUTES,
   PROP_NO_SHADOW,
   PROP_SHADOW_CLASS,
-  PROP_OBSCURED
+  PROP_OBSCURED,
+  PROP_OPACITY
 };
 
 #define DEFAULT_SHADOW_RADIUS 12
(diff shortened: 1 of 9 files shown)


clefebvre:
I'm confused, I thought this was already fixed in 448? I can modify window opacity with the mouse wheel on the titlebar here on master... 448 fixed that.

jaszhix:
No, it doesn't fix that, it just doesn't regress it like #437 (or reverted commits?) does - as discussed on Slack.

This fixes opacity during effects. From @itzexor in linuxmint/cinnamon#8454:

> simple test case:
> - WinKey + E to open nemo, WinKey + L to open melange
> - in melange start inspector and click the nemo window
> - evaluate r(#).opacity = 127 where # is the result number of the nemo window inspection

src/compositor/meta-window-actor.c
```diff
@@ -3580,41 +3359,96 @@ meta_window_actor_invalidate_shadow (MetaWindowActor *self)
   clutter_actor_queue_redraw (CLUTTER_ACTOR (self));
 }
 
+static inline guint8 OPACITY_TYPE_CARDINAL = -1;
```
mtwebster: This will not work.  This ends up being 255 when you try to assign it to an unsigned int.  Which ends up causing line 3376 to evaluate true any time you're at full opacity (255).

Open a nemo window, activate the debugger (ctrl-shift-d), pick the NemoWindow properties, and assign the opacity to something less than 1.0.  The window updates.  Now use the titlebar-scrolling to restore the opacity.  It gets into a tug-of-war with the window opacity atom.

I think you have to either make set_opacity take a gint argument, then cast it to guint (eg https://github.com/mtwebster/muffin/commit/3835e03e6948a8edc82ca75471b3c6b941162de0) or make a separate method for the odd codepath that potentially reads the atom before calling set_opacity (https://github.com/mtwebster/muffin/commit/a68c41db1fbc727ecec9915104b3ae743f2164c3 minus the property stuff).  I prefer the latter, it's much simpler to understand (and maybe more efficient?)

Otherwise it seems to manage the opacity properly without issue, reacts to client opacity changes (and restores it after a restart).

PR #450, opened 2019-03-29
window: Fix the snap OSD getting stuck

jaszhix:

### DIFF
--- a/src/core/window.c
+++ b/src/core/window.c
@@ -9368,6 +9368,7 @@ update_move (MetaWindow  *window,
     return;
 
     gboolean last_tile_mode_state = window->tile_mode;
+    gboolean last_mouse_on_edge_state = window->mouse_on_edge;
 
   /* Originally for detaching maximized windows, but we use this
    * for the zones at the sides of the monitor where trigger tiling
@@ -9571,7 +9572,8 @@ update_move (MetaWindow  *window,
   gboolean hminbad = FALSE;
   gboolean vminbad = FALSE;
 
-  if (window->tile_mode != last_tile_mode_state)
+  if (window->tile_mode != last_tile_mode_state ||
+      window->mouse_on_edge != last_mouse_on_edge_state)
     {
       if (meta_prefs_get_edge_tiling ())
         {
--- a/src/core/window.c
+++ b/src/core/window.c
@@ -9367,6 +9367,8 @@ update_move (MetaWindow  *window,
   if (dx == 0 && dy == 0)
     return;
 
+    gboolean last_tile_mode_state = window->tile_mode;
+
   /* Originally for detaching maximized windows, but we use this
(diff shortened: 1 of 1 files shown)


JosephMcc:
Ok, I gave this a quick test. I can no longer reproduce the stuck OSD but it brings new problems. If I move to one side of the screen that pulls up the small tile zone indicator but not far enough to bring up the actual tile preview, the tile preview will not show if I move all the way to the edge.

In addition now when moving directly to the edge I get both the small indicator and the tile preview.

mtwebster:
For posterity:


i see the same thing @joseph noticed in the 450 patch, i noticed this commit https://github.com/linuxmint/muffin/commit/9b849444303c72f14bd1c68b104db1a01ff56139 - which is from 4.0 (i dont even know if this commit is relevant, but it does concern the previews/hud).  i'll play around some more tomorrow.

mtwebster [10:25 PM]
i tried master with that reverted, it seems identical to the current behavior (stuck text window), which i guess it was intended to fix in the first place (and must have been occurring already in 4.0 then)
ok tomorrow

mtwebster [10:30 PM]
ok one thing to state, the current master behavior (other than the stuck text window) is 'correct' - you get the little edge blob when you get close, and the full window-sized preview once you reach the edge (with the blob fading out as well).
the only thing i don't remember - and either way i don't care, is if the full preview was over or under the grabbed window
the blob appears over the window, but the preview slides out under.  i don't remember if that's different or not, but I think it's ok as it is

JosephMcc:
I'm fairly certain at least at one point the preview was below the window. At least it was intended to be that way.

clefebvre:
I can confirm the tile preview (i.e. the big one) should be behind the window, and it is. Arguably the little one (the hud preview) could be behind as well (it's on top on master, and I think it was on top in the past as well), I don't think that needs to be fixed or worried about in the scope of this PR though.

clefebvre:
I see the same issues as @mtwebster and @JosephMcc as well.

jaszhix:
In Cinnamon, I replaced the deprecated `raise_top` method with `parent.set_child_above_sibling`.

https://github.com/linuxmint/cinnamon/commit/394b97c253f166dd060359e993a7f912b9fea4b4#diff-3411cdb80aaaddfcaa2c6d036a436b69L147

If it was broken before, I just fixed it to work as originally intended. I use tiling every day, and don't see any regressions from this PR.

jaszhix:
If what you're talking about is the tile HUD not popping up as quickly (and it still pops up before the window reaches the edge for me), we should really determine if that's something we need and users want. This PR improves window dragging smoothness for me, something I know @JosephMcc has been complaining about since I started touching muffin. I can't really make the HUD appear any sooner without calling into JS more, which makes the elastic effect on window dragging worse.

PR #451, opened 2019-03-29
window-actor: Only toggle sync for fullscreen apps while redirected

jaszhix:
This fixes an issue with windows freezing in some situations. To reproduce, disable unredirection for fullscreen windows, and launch cinnamon-screensaver with d-feet (Session Bus -> org.cinnamon.ScreenSaver -> SetActive). This doesn't reproduce in normal situations for me, but @itzexor has run into it on amdgpu.

### DIFF
--- a/src/compositor/meta-window-actor.c
+++ b/src/compositor/meta-window-actor.c
@@ -2264,6 +2264,17 @@ meta_window_actor_should_unredirect (MetaWindowActor *self)
   return FALSE;
 }
 
+static void
+fullscreen_sync_toggle (MetaWindowActor *self,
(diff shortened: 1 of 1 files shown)


itzexor:
This doesn't seem to fix the redraw freeze on sync change for me. A near 100% reproducer for me is to have MPV fullscreen on one monitor, ensure the sync method has switched by having a terminal tail the session log, then click the menu applet on the second monitor. Once the menu comes up it often freezes for ~30s

jaszhix:
Pushed a commit to address this in #452, which includes this patch.

PR #463, opened 2019-04-07
window-actor: Use int for opacity

jaszhix:
This fixes an issue with #449 where client requests didn't propagate due to the max possible value for `guint8` being 255.

### DIFF
--- a/src/core/window.c
+++ b/src/core/window.c
@@ -4333,7 +4333,7 @@ meta_window_adjust_opacity (MetaWindow   *window,
 {
   MetaWindowActor *actor = META_WINDOW_ACTOR (window->compositor_private);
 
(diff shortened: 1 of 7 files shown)


src/core/window.c
```diff
@@ -4333,7 +4333,7 @@ meta_window_adjust_opacity (MetaWindow   *window,
 {
   MetaWindowActor *actor = META_WINDOW_ACTOR (window->compositor_private);
 
-  gint current_opacity, new_opacity;
+  int current_opacity, new_opacity;
```
mtwebster: these are exactly the same fwiw (see [here](https://developer.gnome.org/glib/stable/glib-Basic-Types.html#glib-Basic-Types.description))

PR #468, opened 2019-04-08
cogl-pipeline-layer-private.h: remove a redundant redeclaration

brownsr:
and lose 800 lines of build spam


jaszhix:
Can you please clarify where this build spam is coming from? If it is originating from muffin (in src/), we should fix the issues there. It's difficult to maintain a diverging code base this large, so I don't think we should touch these unless we absolutely have to. These changes may look simple to us, but on a low level, GCC optimizations can change from small details like this.

collinss:
I don't think this can be 'fixed' in muffin itself without removing the imports themselves, and even that wouldn't truly be a fix. However, if you removed `-Wredundant-decls` from the compiler flags for muffin, it would silence the warnings at least. The reason these warnings don't show up when compiling cogl is that that flag isn't supplied.

PR #469, opened 2019-04-08
cogl-pipeline-private.h: remove a couple of redundant declarations

brownsr:
and say goodbye to nearly 1100 lines of build log warnings

### DIFF
--- a/cogl/cogl/cogl-pipeline-private.h
+++ b/cogl/cogl/cogl-pipeline-private.h
@@ -846,9 +846,6 @@ _cogl_pipeline_weak_copy (CoglPipeline *pipeline,
 void
 _cogl_pipeline_set_progend (CoglPipeline *pipeline, int progend);
 
-CoglPipeline *
-_cogl_pipeline_get_parent (CoglPipeline *pipeline);
-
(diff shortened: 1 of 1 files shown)


cogl/cogl/cogl-pipeline-private.h
```diff
@@ -924,10 +921,6 @@ _cogl_pipeline_set_static_breadcrumb (CoglPipeline *pipeline,
 unsigned long
 _cogl_pipeline_get_age (CoglPipeline *pipeline);
 
-CoglPipeline *
-_cogl_pipeline_get_authority (CoglPipeline *pipeline,
```
jaszhix: I would recommend not diverging from Mutter too much in Cogl/Clutter. Mainly, it makes cherry-picking simpler in the future. But also, I'm experiencing perf regressions from these changes, and I don't know why. In this file for instance, `_cogl_pipeline_get_authority` is inlined, and the second definition exports it. 

I'm also not seeing any warnings in Clutter and Cogl from the original upstream changes, so I would appreciate it if you could help me understand what warnings this is fixing.

PR #473, opened 2019-04-10
frames.c: add prototype, remove unused variable

brownsr:


src/ui/frames.c
```diff
@@ -28,6 +28,7 @@
 #include <string.h>
 #include <meta/boxes.h>
 #include "frames.h"
+#include "frame.h"
```
jaszhix: This change has implicit performance side effects. MetaFrame is passed as an argument in this file, like with

```
flags = meta_frame_get_flags (frame->meta_window->frame);
```
but we don't need to include its header unless we're accessing a property off its struct. Since frames.c is compiled much later, the function symbols are available. The structs primarily used here are MetaFrames and MetaUIFrame.

I don't understand why, but I tried this a while ago and it decreased performance. It might be an introspection quirk. Testing this patch, my previous experience was confirmed. Mutter doesn't include this header either, and it has the same meta_frame_* calls in frames.c.

Another way to solve this if it's causing a warning might be to just move the function definitions used in this file to window-private.h, which is included in frame.h.

brownsr:
OK, closing.

PR #483, opened 2019-04-21
4.0-maintenance: Revert "Phase out MetaCompScreen (#424)"

jaszhix:
This commit slipped by my radar because #405 was open for a while, but this commit is not stable yet and is causing Cinnamon to segfault during restarts.


jaszhix:
#507

PR #485, opened 2019-04-22
window.c: Remove a condition for updating the outer rect

JosephMcc:
We only update the outer rect if the edge resistance setting is enabled. Since
this setting is on by default it's being called on my most users installs
anyway. This also has the side effect of fixing other things that depend on
the rect being updated. An example of this is the workspace switcher Cinnamon
applet.


JosephMcc:
Lol, sorry. Totally missed this: https://github.com/linuxmint/muffin/pull/484

PR #489, opened 2019-04-23
Fix screensaver regressions

jaszhix:


mtwebster:
This seems to still always fail the first time the screensaver is activated.  Subsequent activations appear to work (with current master cinnamon-screensaver).

I've been messing around with the screensaver fade properties as well.

https://github.com/linuxmint/cinnamon-screensaver/blob/master/src/constants.py#L7-L8

If I set the fade in/out to > 500 it doesn't always work.  If I set it substantially higher (like 1000), the screensaver never activates properly.

On the other hand, if you set those to 0 (which skips any fading) the screensaver always works properly.  I was considering committing this just to patch this problem for the time being, but even with this patch I still occasionally see the root window briefly immediately after deactivation.  Without this patch it's even worse.

And of course my worry about simply removing the fade from the screensaver is that some other case will come up for some other program none of us has tried, so I'm not sure what to do here.

jaszhix:
Ah, I'm not sure either. I was curious what other applications might end up using the `META_WINDOW_OVERRIDE_OTHER` window type. It's definitely specific to OR windows, but most fullscreen OR windows do not end up in the top window group - override_other types are filtered into the `top_window_group`. I'm suspecting some code paths are treating `top_window_group` OR windows as if they're in the normal window group, and causing this paint issue. I should have time to give this another look.

jaszhix:
The fade is indeed the trigger. With this patch I do see the frame rate chug a bit with fade in use, but it does animate where master doesn't. 

It's not normal for applications to be in the top window group - if you poll `global.top_window_group.get_children()` in melange, you'll see it's only cinnamon, cinnamon-screensaver, or popup windows. These are windows that need to be treated as chrome basically, and the screensaver seems to be the only case where it would be chrome and unredirected.

I recommend we remove the fade from the screensaver and apply this patch, and if more is needed to be done we can go further. Another approach might be setting different window hints - one possibly might be setting the window as sticky instead of using the `META_WINDOW_OVERRIDE_OTHER` window type in order to keep the window on top. As far as Clutter is concerned the end result is the same, all these windows are on the stage, but stacked in a different order.

jaszhix:
Do you think the fade might also fix this?

https://github.com/linuxmint/cinnamon-screensaver/issues/304

mtwebster:
This seems to take care of the screensaver issue with and without the fades disabled, as well as the problem with the desktop being shown initially after deactivation.

PR #491, opened 2019-05-01
window.c: various minor tweaks, small speedup

brownsr:
move code to the far side of early return statements where possible, change scope of a variable
declaration, a few small things to avoid doing unnecessary work.  Benchmarks as a 0.3% improvement in full screen window
performance (furmark)


jaszhix:
Can't see a difference over master, but not seeing anything wrong with this either except not initializing pointers with `NULL`.

PR #492, opened 2019-05-14
compositor: fix a few compiler warnings

brownsr:


src/compositor/compositor.c
```diff
@@ -1250,6 +1243,9 @@ meta_post_paint_func (gpointer data)
       clutter_actor_queue_redraw (CLUTTER_ACTOR (compositor->stage));
       break;
 
+    case COGL_GRAPHICS_RESET_STATUS_GUILTY_CONTEXT_RESET:
```
jaszhix: I know GCC complains about this, but what is it solving? Is there a performance hit with checking more cases vs implicit? We _can_ configure our build to not warn over this, and it seems like an opinion to me.

brownsr: Yes, I think I agree.  removed these.

PR #493, opened 2019-05-14
boxes: remove an ISO C90 warning

brownsr:


jaszhix:
Slightly off topic - this applies to all the warning cleanup PRs in general, and it's not necessarily meaning something is wrong with them on their own - one thing that is worth noting is if these are C90 warnings, it appears the newest revision of C is C18. Languages generally evolve, and so do their compiler optimizations.

> The same standard as C89 was ratified by the International Organization for Standardization as ISO/IEC 9899:1990, with only formatting changes,[2] which is sometimes referred to as C90.

https://en.wikipedia.org/wiki/ANSI_C#C90

I really do think we should evaluate which warnings are stylistic and which are actually preventing GCC from optimizing correctly because 29 years is a lifetime in computer science. I have seen cases first hand from my own attempts at warning clean up, where the suggested change ended up hurting performance. It's worth considering for every change.

brownsr:
Agree these are hardly the most consequential thing.  I admit I do find variables declared in the middle of a block of code less readable, so to that extent the C90 warnings reflect that.  I suppose I could read them as a little nudge by the compiler to think about restructuring as well.  Whether optimisation is affected, I am afraid I don't know.  I'd rather doubt it for something as simple as this.

jaszhix:
Yeah, that comment is not related to this PR (which is good, this would fail in the clutter build because of maintenance flags), but it's kind of a summary of my concerns after looking through these round of PRs. However, it does make me wonder if we're sacrificing something for readability. In most projects I don't prioritize performance to this degree, but since I interact with all my other apps through muffin, I've been looking into these little things more. Many, many gotcha's in C:

http://leto.net/docs/C-optimization.php

brownsr:
Understood.  But C code can run unbelievably fast, so if something feels slow the problem is usually an algorithm logic fault, disk io, inter-process communication or something else inherently slow that the computer is waiting for.

jaszhix:
I wish I was just force feeding myself a steady stream of placebo misinterpretations of responsiveness so I wouldn't worry as much to be honest. Most of the time I check my assumptions with benchmarks though, they're validated. The article I linked is kind of old, but makes a good point about using common sense and situational awareness, which is not always intuitive until you've been modifying the same few C files for months on end. :slightly_smiling_face:

PR #495, opened 2019-05-16
gradient: resolve a few compile warnings

brownsr:


src/ui/gradient.c
```diff
@@ -861,5 +867,8 @@ meta_gradient_add_alpha (GdkPixbuf       *pixbuf,
     case META_GRADIENT_LAST:
       g_assert_not_reached ();
       break;
+    default:
+      g_warning("reached default in meta_gradient_add_alpha");
```
jaszhix: Why do we need to warn over this? The default operator addition looks fine, GCC probably adds it implicitly and that's why it's complaining. That doesn't mean there's something wrong at _runtime_, it's just complaining about inefficient compilation, which doesn't affect users.

brownsr: OK, removed the warning.  I must admit the asserts in the code do worry me, as they will crash if the code is built without disabling them, and we can't control how all distros will build our code.  If we ever get to Wayland then I imagine that will be a full session crash, not just a Cinnamon restart needed.  I find the thought of deliberately crashing a DE bizarre, I personally would incline to a suitable warning message and carry on.

PR #498, opened 2019-05-18
meta-window-group: resolve some compile warnings

brownsr:


src/compositor/meta-window-group.c
```diff
@@ -86,7 +86,7 @@ meta_window_group_cull_out (MetaWindowGroup *group,
 
   while (iter_prev (&iter, &child))
     {
-      if (!CLUTTER_ACTOR_IS_VISIBLE (child) ||
+      if (!clutter_actor_is_visible (child) ||
```
jaszhix: Compare these two:

```c
#define CLUTTER_ACTOR_IS_VISIBLE(a) \
  CLUTTER_MACRO_DEPRECATED_IN_1_24_FOR ("Deprecated macro. Use clutter_actor_is_visible instead") \
  ((((ClutterActor*)(a))->flags & CLUTTER_ACTOR_VISIBLE) != FALSE)
```
```c
gboolean
clutter_actor_is_visible (ClutterActor *self)
{
  g_return_val_if_fail (CLUTTER_IS_ACTOR (self), FALSE);

  return CLUTTER_ACTOR_IS_VISIBLE (self);
}
```
This is a (premature?) de-optimization because `CLUTTER_ACTOR_IS_VISIBLE` is the inline macro that is just using a bit operation. No function call in its current form, but this adds one. We need to decide if a deprecation makes sense for muffin, which is not a dynamically linked library to Clutter. I added Clutter's private headers to the compositor code, so we can pretend to be inside Clutter most of the time and escape the rules meant for third party applications. This also gives GCC more optimization hints because the private structs are exposed to the compositor.

brownsr: OK

PR #500, opened 2019-05-18
constraints: remove some compile warnings

brownsr:


src/core/constraints.c
```diff
@@ -829,15 +827,17 @@ constrain_maximization (MetaWindow         *window,
       if (g_list_length (window->screen->active_workspace->snapped_windows) > 0) {
         GList *tmp = window->screen->active_workspace->snapped_windows;
         GSList *snapped_windows_as_struts = NULL;
+        MetaStrut *strut;
+        MetaSide side;
```
jaszhix: This implicitly initializes a pointer to a random address. While this may be harmless because we're definitely using it here, I think it's good practice to assign NULL. Especially when the code is complex with early returns, it's nice to know we can sanity check any pointer and know we're not referencing something random in memory, and that we're not creating overhead pointing to a random address and then not using the pointer by returning early.

https://www.acodersjourney.com/top-20-c-pointer-mistakes/

brownsr: OK, done

brownsr:
The merge conflicts are pretty awkward, so too much chance of introducing an error as this is simply a warning fix.  Closing.  Can come back to this sometime.

PR #504, opened 2019-05-19
window-props: remove some compiler warnings

brownsr:


src/core/window-props.c
```diff
@@ -277,7 +277,7 @@ reload_progress (MetaWindow    *window,
       if (window->progress != 0)
         {
           window->progress = 0;
-          g_object_notify (window, "progress");
+          g_object_notify (G_OBJECT(window), "progress");
```
jaszhix: I would recommend:

```c
g_object_notify ((GObject*) window, "progress");
```
because the instance cast function might be more verbose when it's not a GObject, but they inline layers of macros when this should be a simple cast, a relatively cheap operation. If we're this deep into the code, we know this MetaWindow is a GObject or muffin would crash, and we have nothing to gain from the extra complexity here.

brownsr: OK, done

jaszhix: Thanks

PR #505, opened 2019-05-25
[Next] [RFC/WIP]buffer and frame timing changes from upstream

brownsr:
This has the buffer and frame timing changes from upstream, that have been fairly widely trailed [phoronix etc.].  I'm not sure I can notice the difference, but Daniel van Vugt gives reasons why the problem was not noticed here https://gitlab.gnome.org/GNOME/mutter/commit/45244852acc214a7a9d01dc96896ad0c079b437c

so perhaps those reasons apply for us. Anyway I know I am not as observant as others over frame timing issues.  Here to try out.  Not to pull.

### DIFF
--- a/clutter/clutter/cogl/clutter-stage-cogl.c
+++ b/clutter/clutter/cogl/clutter-stage-cogl.c
@@ -166,84 +166,56 @@ clutter_stage_cogl_schedule_update (ClutterStageWindow *stage_window,
 {
   ClutterStageCogl *stage_cogl = CLUTTER_STAGE_COGL (stage_window);
   gint64 now;
-  gint64 target_presentation_time;
+  float refresh_rate;
   gint64 refresh_interval;
-  /*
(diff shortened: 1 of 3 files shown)


jaszhix:
This seems to make frame variability more pronounced, but it's subtle. I tested a version of this commit a few months ago and noticed the same thing. I was going to look at this again when [this MR](https://gitlab.gnome.org/GNOME/mutter/merge_requests/363) was merged. Since we support disabling sync we have custom frame time handling in `clutter_stage_cogl_schedule_update`, as well as in clutter-master-clock-default.c.

We also support disabling sync, so that is something to keep in mind when testing frame timing patches from Mutter. I don't think we'll be able to follow Mutter as close as we need to be to introduce this patch without side effects, without also removing the [sync timing changes](https://github.com/linuxmint/muffin/commit/42f9e08b361f34a15494969b637e483cdd435063).

PR #510, opened 2019-06-04
Fix screensaver rendering issues introduced by unredirection bug fixes

jaszhix:
This forces compositing on the screensaver. I suspect this might be related to upstream refactoring that fixed issues with OR windows not being fully unredirected when they should be. The screensaver sets the `META_WINDOW_OVERRIDE_OTHER` window type so it is added to the `top_window_group` and always above all windows. This seems to have an issue with the actual underlying unredirection with Xorg though, and suggests the screensaver may not have been unredirecting before until 63ddb021f2393610e16f44949e9890fc86e3ad51 - where we see the render updates freezing. Not exactly sure why this occurs, but after testing both native and wine games, so far the screensaver is the only app I've noticed having rendering issues when unredirected.

Closes https://github.com/linuxmint/alpha-testing/issues/10

### DIFF
--- a/src/compositor/compositor-private.h
+++ b/src/compositor/compositor-private.h
@@ -55,6 +55,8 @@ struct _MetaCompositor
   guint           show_redraw : 1;
   guint           debug       : 1;
   guint           no_mipmaps  : 1;
+  // Screensaver case
+  guint           override_window_on_top;
 
   gboolean frame_has_updated_xsurfaces;
   gboolean have_x11_sync_object;
--- a/src/compositor/compositor.c
+++ b/src/compositor/compositor.c
@@ -1122,7 +1122,22 @@ get_top_visible_window_actor (MetaCompositor *compositor)
       meta_window_get_input_rect (meta_window_actor_get_meta_window (actor), &rect);
 
       if (meta_rectangle_overlap (&compositor->display->active_screen->rect, &rect))
-        return actor;
+        {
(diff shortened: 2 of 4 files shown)


jaszhix:
Rebased. Please read the commit comments. I accidentally rebased on to @brownsr's remote initially, but fixed it.

jaszhix:
Clem has told me on Slack that he thinks I should take a break, even though I'm just trying to understand why cinnamon is more broken now than it was before. Also, he thinks the fact I am saddened by this is me feeling sorry for myself, and has completely ignored my attempts to get some kind of acknowledgement about a change that was pushed untested. He thinks this PR is a "workaround", even though the screensaver was **never** unredirecting in 4.0 and is only broken now because the mutter patches fixed it. So I guess 4.2 will not have the improvements for gaming. I no longer have confidence the project is heading in the right direction and will be spending my time where it is more appreciated.

I want a basic level of respect and dignity, as everyone deserves. Closing in case it gets merged and people want me to support it and work with someone who has unapologetically demonstrated extreme rudeness and contempt, and somehow thinks I would come back to that work environment.

jaszhix:
I would delete this PR if I could. That's how not welcome to my work you guys are, only willing to extend what I'm legally required to because of the GPL. You guys don't deserve anything. Revert all my work and then asked me to leave the team because BDFL can't have anyone disagreeing with him. I'll happily leave a dictatorship.

PR #521, opened 2019-06-10
[Next] 4.0: Don't set CLUTTER_ACTOR_NO_LAYOUT flag

clefebvre:
Fixes the flydown issue in https://github.com/linuxmint/alpha-testing/issues/2.

This could also be responsible for flickering regressions in upgrades from 3.8 to 4.0 (where this flag is set in init())
and from 4.0.6 to 4.0.7 (where it's set in init and removed during window grabbing).

Build for 4.0-maintenance:
https://github.com/linuxmint/cinnamon/files/3273029/4.0-maintenance-no-norelayout-flag.zip

### DIFF
--- a/src/compositor/compositor-private.h
+++ b/src/compositor/compositor-private.h
@@ -77,9 +77,6 @@ void     meta_end_modal_for_plugin   (MetaScreen       *screen,
 gint64 meta_compositor_monotonic_time_to_server_time (MetaDisplay *display,
                                                       gint64       monotonic_time);
 
-void meta_compositor_grab_op_begin (MetaCompositor *compositor);
-void meta_compositor_grab_op_end (MetaCompositor *compositor);
-
 void meta_check_end_modal (MetaScreen *screen);
 
 #endif /* META_COMPOSITOR_PRIVATE_H */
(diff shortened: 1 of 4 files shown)


jaszhix:
> This could also be responsible for flickering regressions in upgrades from 3.8 to 4.0

This flag was not introduced until 4.0.

clefebvre:
I know, that's why it can affect people upgrading from 3.8 to 4.0.

clefebvre:
potentially https://github.com/linuxmint/cinnamon/issues/8613

jaszhix:
I think you might be putting too much stock in every little fix. There's no magic bullet, at a certain point the whole architecture just has a ton of technical debt. I did what I could, but don't use me as a scapegoat for all of cinnamon's existing issues. A lot of this was inherited from GNOME. This flag was only responsible for windows disappearing for drag. The fly up effect was new to me, but also doesn't exist in GNOME. Definitely not related to flicker at all, that's as old as cinnamon 2.x.

clefebvre:
I don't think like that. There's no scape goat and there's no finger pointed here. I couldn't care less if the issue comes from you or from gnome. We're all trying our best, you, me, them. If reverting this fixes that, then it's worth considering. This is not about you.

jaszhix:
Well, it's absurd to think something that always existed in Cinnamon is caused by something I recently added.

jaszhix:
In fact, the obscured optimization fixed flickering. yes, this is something I really don't want my _work_ (an inanimate thing, not _me_) blamed on.

PR #526, opened 2019-07-06
clutter: Deliver events sooner when possible.

clefebvre:
From @vanvugt
https://gitlab.gnome.org/GNOME/mutter/commit/ae8fc6146b14c1905a99a0e929d1f192f2ed050a

Previously all events would be queued and their processing deferred till
the next master clock tick, at which point supersesed input events would
be dropped and only the latest of each type used. This was great for
minimizing CPU usage but had two drawbacks:

 * Clients would receive the next input event after it is already too
   late to make it to the next compositor frame.
 * Clients would receive a lower resolution event stream than the hardware
   is capable of.

We now instead scale performance dynamically according to available time.
If there is enough idle time available then that will be used to deliver
events immediately without delay. Otherwise event delivery will scale down
to the old minimal-CPU behaviour.

This allows clients to receive input events sufficiently in advance of the
next compositor frame that they can respond and redraw with one frame
lower latency than before. It also allows clients higher resolution input,
in case they are able to use it.

### DIFF
--- a/clutter/clutter/clutter-stage.c
+++ b/clutter/clutter/clutter-stage.c
@@ -122,6 +122,7 @@ struct _ClutterStagePrivate
   ClutterActor *key_focused_actor;
 
   GQueue *event_queue;
+  guint event_flushing_idle_source;
 
   ClutterStageHint stage_hints;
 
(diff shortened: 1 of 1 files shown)


clefebvre:
Build (on top of 4.2): [muffin.zip](https://github.com/linuxmint/muffin/files/3364572/muffin.zip)

PR #527, opened 2019-07-08
Add function to fetch a window's pid from XCB

Cobinja:
This works more reliable for windows from flatpaks than the usual net_wm_pid.


Cobinja:
Force pushed a version where the client's pid is only fetched once, not on every call to get it

PR #533, opened 2019-08-21
[testing only] Mutter frame timing rebase

itzexor:
This is just for anyone who wants to try the recent changes in mutter to frame timing. It also includes the proposed nvidia-specific threaded-swap removal which should improve performance.

This reverts a ton of local changes, and as a result **will break cinnamon-settings** while applied since some gsettings are removed, and I haven't created a complementary branch for cinnamon.

The gist of the changes here are:
1. Revert all local changes that affected frame timing code paths so we are basically at the code as it existed when mutter's fork of clutter/cogl were merged into muffin.
2. Apply the following mutter merge requests:
   1. [Remove ability to externally set sync-to-vblank](https://gitlab.gnome.org/GNOME/mutter/merge_requests/191)
   2. [clutter/stage-cogl: Don't skip over the next frame](https://gitlab.gnome.org/GNOME/mutter/merge_requests/520)
   3. [clutter/stage-cogl: Reschedule update on present](https://gitlab.gnome.org/GNOME/mutter/merge_requests/281)
   4. [Consolidate all frame throttling into clutter-stage-cogl](https://gitlab.gnome.org/GNOME/mutter/merge_requests/363)
   5. [cogl: Remove GLX "threaded swap wait" used on Nvidia](https://gitlab.gnome.org/GNOME/mutter/merge_requests/602)
   6. [Remove pending_swaps counter](https://gitlab.gnome.org/GNOME/mutter/merge_requests/719)


mfowen:
> This is just for anyone who wants to try the recent changes in mutter to frame timing

Except the author of the commits you're reverting, who is banned, and can't provide feedback.

PR #540, opened 2019-11-21
meta-shadow-factory: Fix terminal shadows

clefebvre:
The regression was introduced in https://github.com/linuxmint/muffin/commit/66efde884efce35afd433efbda966774a65c9d85

The rectangle array wasn't being filled up properly.

Screenshot of the issue:

image


leigh123linux:
It also affected firefox and others.
I can confirm it fixes the regression.

PR #541, opened 2019-11-21
window.c: Fix regression on window move

clefebvre:
In https://github.com/linuxmint/muffin/commit/89805c059609f2cd48f60148896c057428362cb6
it looks like two lines were removed by mistake, making it impossible to
move windows.


leigh123linux:
I can confirm it fixes the regression

PR #547, opened 2019-12-20
[WIP] Catch up clutter/cogl a bit with upstream to ease the transition to meson

collinss:


brownsr:
A first try at a build appears to jam up in the build, the last message was PASS: texture 1 /texture/pick-with-alpha  The screen went white, I had to alt-tab back, that white window is titled events-touch.  I'm not on a touch-capable screen, so perhaps it is sat waiting for a touch event test it can never get.

collinss:
@brownsr Thanks for testing! I've tried this build on 2 different machines now, and neither has a touch screen, and they worked fine. But who knows, perhaps it's one of those heisenbugs. I seem to remember muffin failing the tests randomly in the past, so it could just be more of that.

mtwebster:
I can't build this with dpkg-buildpackage - there are all sorts of missing/misplaced files.  I made a little start here:

https://gist.github.com/mtwebster/eb0278b55f000744d1079fb7a31a518a

You probably should make a copy of the tree, build master, and compare the debian/tmp tree between them - the files are all being generated I think, they're just not where the debian/*.install files are looking, and I'm pretty sure some of the locations are wrong (but I'm no expert).

Did you pull these in from gnome?  Maybe have a look at ubuntu's source tree for mutter?  Their quilt patches might give some insight - https://code.launchpad.net/ubuntu/+source/mutter.  For each branch there there's one with the patches applied already, and one without.  I'm not certain how different the bionic ones are from focal (our next base), but probably focal is more important, as we don't usually backport cinnamon.

collinss:
@mtwebster I'm not sure what happened - I know I built it with dpkg before and it worked just fine. I dropped one of the commits that was just moving things around, and that took care of most of it, but I was still getting extra libmuffin-clutter-0.so.blah stuff. I still have no idea what change here could have possibly caused that - nothing seems related, but I did manage to fix it. I also rebased onto master.

PR #548, opened 2020-01-24
fix: cogl-texture-2d.c:432:43: error: 'EGL_WAYLAND_BUFFER_WL' undeclared (first use in this function); did you mean 'EGL_TRIPLE_BUFFER_NV'?

mkg20001:
```
cogl-texture-2d.c:432:43: error: 'EGL_WAYLAND_BUFFER_WL' undeclared (first use in this function); did you mean 'EGL_TRIPLE_BUFFER_NV'?
                                             EGL_WAYLAND_BUFFER_WL,
                                             ^~~~~~~~~~~~~~~~~~~~~
                                             EGL_TRIPLE_BUFFER_NV
```

  https://bugs.gentoo.org/698736 - Muffin is a mutter fork, so patch _should_ apply

### DIFF
--- a/cogl/configure.ac
+++ b/cogl/configure.ac
@@ -487,7 +487,8 @@ AS_IF([test "x$enable_gles1" = "xyes"],
 
             AC_CHECK_HEADERS([EGL/eglext.h],
                              [COGL_EGL_INCLUDES="$COGL_EGL_INCLUDE
-#include <EGL/eglext.h>"],
+#include <EGL/eglext.h>
+#include <EGL/eglmesaext.h>"],
                              [],
                              [$COGL_EGL_INCLUDES])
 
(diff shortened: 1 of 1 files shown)


leigh123linux:
muffin should be compiled without wayland support so I doubt this change is needed.

https://github.com/linuxmint/muffin/blob/master/debian/rules#L34

worldofpeace:
> muffin should be compiled without wayland support so I doubt this change is needed.
> 
> https://github.com/linuxmint/muffin/blob/master/debian/rules#L34

Not sure I understand this. If you have configure options that you believe will always break the build why have it there? And if this is complied with wayland support this is clearly a bug that could be fixed.

mtwebster:
https://github.com/linuxmint/muffin/commit/f78bf5b309b3d306848f47cc241b31e9399999a7 - I'd rip all that code out of there if it wasn't a waste of time.  If we support wayland some day this won''t stand in our way I think :)

worldofpeace:
> [f78bf5b](https://github.com/linuxmint/muffin/commit/f78bf5b309b3d306848f47cc241b31e9399999a7) - I'd rip all that code out of there if it wasn't a waste of time. If we support wayland some day this won''t stand in our way I think :)

Thank you @mtwebster :sparkles:

PR #549, opened 2020-02-05
pkg-config: support cross-build cinnamon

sgn:
On cross-compiling,

    $PKG_CONFIG --variable=girdir libmuffin

returns `/usr/lib/muffin` instead of `$sysroot/usr/lib/muffin`

From: https://autotools.io/pkgconfig/cross-compiling.html

> To design a variable to contain a path that needs to be used at build
> time, such as the path where a generation script is, you can prefix it
> in the .pc file with the built-in variable ${pc_sysrootdir}.

Fix it by employing advise from autotools.

---

related: https://github.com/void-linux/void-packages/pull/18826


eli-schwartz:
Hmm, I notice ever since this change I'm getting an rpath for /usr/bin/cinnamon which contains the pc_sysrootdir, which is "/", so

```
$ readelf -d /usr/bin/cinnamon| grep rpath
 0x000000000000000f (RPATH)              Library rpath: [/usr/lib/cinnamon://usr/lib/muffin]
```

`//usr/lib/muffin` seems a bit interesting here. :p

leigh123linux:
+1 for reverting 

```
[leigh@leigh ~]$ readelf -d /usr/bin/cinnamon| grep rpath
 0x000000000000000f (RPATH)              Library rpath: [/usr/lib64/cinnamon://usr/lib64/muffin]
```

sgn:
On 2020-05-22 14:10:49-0700, Eli Schwartz <notifications@github.com> wrote:
> Hmm, I notice ever since this change I'm getting an rpath for /usr/bin/cinnamon which contains the pc_sysrootdir, which is "/", so
> 
> ```
> $ readelf -d /usr/bin/cinnamon| grep rpath
>  0x000000000000000f (RPATH)              Library rpath: [/usr/lib/cinnamon://usr/lib/muffin]
> ```
> 
> `//usr/lib/muffin` seems a bit interesting here. :p

I'm fine with reverting this change.

I'll find a proper method to cross-compile cinamon later.
I couldn't recall why I need to prepend ${pc_sysrootdir} though.

-- 
Danh

PR #576, opened 2021-01-15
[In Testing] Some deprecation fixes

ItzSwirlz:
It's 10 PM, I'll revisit this tomorrow.

Muffin window demo has been tested, everything else I need to look at and ensure muffin works and won't crash and all.

### DIFF
--- a/src/tools/muffin-window-demo.c
+++ b/src/tools/muffin-window-demo.c
@@ -589,6 +589,10 @@ make_dock (int type)
       break;
     case DOCK_ALL:
       break;
+    default:
+      // Most people prefer docks on the bottom, use as default
+      box = gtk_box_new (GTK_ORIENTATION_HORIZONTAL, 0);
+      break;
     }
 
   window = gtk_window_new (GTK_WINDOW_TOPLEVEL);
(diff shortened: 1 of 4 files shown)


ItzSwirlz:
Just need to test the compositor plugin and using gtk_image_new_from_icon_name and I can take off in testing

src/tools/muffin-window-demo.c
```diff
@@ -636,6 +640,12 @@ make_dock (int type)
       break;
     case DOCK_ALL:
       break;
+    default:
```
mtwebster, line 643: You can put multiple cases that point to the same result..
```
case DOCK_BOTTOM:
case DOCK_ALL:
default:
   gtk_widget_set_size_request. ...
   blah
```
instead of duplicating code.

https://c-for-dummies.com/blog/?p=3607

ItzSwirlz:
I'm actually going to mark as WIP because I found a few other possible issues

ItzSwirlz:
Considering clutter is no longer maintained, and a muffin rebase is on the way, this PR is probably unnecessary.

PR #577, opened 2021-01-26
Fix the build with slibtool.

orbea:
When building muffin with slibtool (https://dev.midipix.org/cross/slibtool) it fails, this PR has two commits to fix the build.

1. Several libtool `.la` files are missing as dependencies in some of the `Makefile.am` files.
2. The build fails with `-no-undefined`, this works with GNU libtool because it silently ignores `-no-undefined` while slibtool does not. This requires building `src/compositor/plugins` separately from the rest of `src/` to ensure `libmuffin.la` exists when it is needed.

Full logs:

[muffin-slibtool.1.log](https://github.com/linuxmint/muffin/files/5871673/muffin-slibtool.1.log)
[muffin-slibtool.2.log](https://github.com/linuxmint/muffin/files/5871687/muffin-slibtool.2.log)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,5 +1,5 @@
 
-SUBDIRS=cogl clutter src po doc data
+SUBDIRS=cogl clutter src src/compositor/plugins po doc data
 
 EXTRA_DIST = HACKING MAINTAINERS rationales.txt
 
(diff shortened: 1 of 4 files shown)


mtwebster:
Thanks

Issue #544, opened 2019-12-07
muffin: doesn't unredirect fullscreen proton-based games


user1-github:
**Issue**
It seems that fullscreen unredirection doesn't work even if "disable compositing for full-screen windows" is enabled in system settings. This causes stuttering in games.

**Steps to reproduce**
Enter any game, make sure vsync is disabled in game and you'll see that there is no screen tearing.

I have attached xsession-errors if you need it.
[xsession-errors.txt](https://github.com/linuxmint/mint19.3-beta/files/3934840/xsession-errors.txt)

Cinnamon version - 4.4.3
GPU and Driver - AMD Radeon RX 580, Mesa 19.3.0-rc6 - kisak-mesa PPA

Original Post: https://github.com/linuxmint/cinnamon/issues/9006

**Update:**
This only affects games running on Steam Proton. Fullscreen unredirection works properly only with native linux games.

**Update 2**
Ok, something is really bizzare about this. Earlier today when I was testing the issue, fullscreen unredirection worked with native linux games, but now it doesn't work with native linux games as well.

pharmville:
For the few steam games I have tried inside cinnamon, the checkbox to "disable compositing for fullscreen windows" doesn't have any effect at all. Maybe instead of detection, just introduce a shortcut key to toggle the compositor on/off, like kwin does. Without a reliable method to disable the compositor, I cannot use gsync inside of cinnamon, because gsync requires the compositor to be off. Therefore, when playing games I am forced to use a different desktop environment, because there doesn't appear to be a way to enable gsync if cinnamon is running. Inside nvidia-settings there is a toggle called "Enable GSYNC visual indicator" so I am using the visual indicator to be able to know when gsync is running successfully or not. No matter what I try, gsync never activates inside of cinnamon for any fullscreen game. That leads me to believe cinnamon is currently incompatible with gsync. I have been able to get gsync working with gnome, kde, and xfce. I havent tried with mate, but I suspect it would work fine with compositor off. TLDR: Feature request - gsync compatibility to games running inside cinnamon.

mtwebster:
We've been thinking about adding the ability to specify per-application whether or not to composite - similar to how in windows, you tweak nvidia settings either globally or specifically (can you still do that? It's been a while).  So at least if something in the automatic detection fails, you can still have what you want.

pharmville:
[...]

PR #590, opened 2021-07-06
compositor.c: Skip offscreen windows when unredirecting.

NikitaKarnauhov:
Offscreen windows may sometimes appear on top of the window
stack even if another fullscreen window is obviously top level.

Fixes https://github.com/linuxmint/muffin/issues/544

See comments in the issue for examples of offending offscreen windows.

### DIFF
--- a/src/compositor/compositor.c
+++ b/src/compositor/compositor.c
@@ -1265,15 +1265,24 @@ meta_pre_paint_func (gpointer data)
   GList *l;
   MetaCompositor *compositor = data;
   GSList *screens = compositor->display->screens;
-  MetaWindowActor *top_window;
+  MetaWindowActor *top_window = NULL;
   MetaWindowActor *expected_unredirected_window = NULL;
 
   if (compositor->windows == NULL)
     return TRUE;
 
-  top_window = g_list_last (compositor->windows)->data;
+  for (l = g_list_last (compositor->windows); l; l = l->prev)
+    {
+      MetaRectangle *rect = &meta_window_actor_get_meta_window (l->data)->rect;
+      if (rect->x + rect->width > 0 && rect->y + rect->height > 0)
+        {
+          top_window = l->data;
+          break;
+        }
+    }
 
-  if (meta_window_actor_should_unredirect (top_window) &&
+  if (top_window != NULL &&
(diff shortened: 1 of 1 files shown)


mtwebster:
I'm not ignoring this, I've just been having trouble reproducing the issue so I can test and confirm before merging - when this bug was first reported I was able to use xcom to test - now that only works in native mode. I'll find something...

NikitaKarnauhov:
> I've just been having trouble reproducing the issue so I can test and confirm before merging

For me the problem was easiest to reproduce when running Divinity: Original Sin 2 Definitive Edition under Proton.

mtwebster:
What version of proton do you use? 6.3-5 works fine with or without this patch for me.

I have a feeling I'll end up just merging this anyhow - I don't like to, but I also don't see much chance of this breaking anything else.

NikitaKarnauhov:
> What version of proton do you use? 6.3-5 works fine with or without this patch for me.

I also tested with Proton 6.3-5 prior to submitting this pull request and the problem was present.

> edit: I'm basing my observations off of the fps overlay that steam provides.

Steam FPS overlay always shows variable values for me (at least when v-sync is disabled).
I have a FreeSync monitor (that shows actual current refresh rate) and also I use G-Sync indicator (that can be enabled in nvidia-settings) to check. So when G-Sync/FreeSync is enabled, and unredirection succeeds, G-Sync indicator is green and the monitor shows variable refresh rate value (that is close to the FPS value that steam overlay shows). And when unredirection fails, G-Sync indicator is grey and the monitor shows constant 144Hz.

mtwebster:
Ok, once I overrode gsync detection in the xserver display config, I got the additional controls to enable the gsync options.

I can confirm this patch does allow gsync to turn on, and flip is shown using the normal nvidia indicator. I never observe any different in fps - around 140fps in divinity using 'ultra' visual settings - or any visual difference. Perhaps because my monitor doesn't actually support it?

With a white desktop background, and all windows minimized, you can see a visual artifact along the right and bottom edges of the monitor (where a normal window's drop shadow typically is), so there's a definite issue with being treated as fullscreen here. This patch corrects that as well.

Issue #596, opened 2021-08-23
tiled CSS classes aren't set on tiled windows


Odyseus:
> muffin version 5.0.1
> Distribution - Linux Mint 20.2
> VirtualBox guest x64

Hello, everybody.

I'm not 100% sure if this is the right place to report this issue. Any advise would be welcome.

### Issue

Styles from Gtk 3 themes for tiled CSD (Client Side Decorated) windows aren't applied. All of Linux Mint Gtk 3 themes have rounded corners for titlebars/headerbars, but when a window is tiled, those corners are removed by targeting the `tiled` CSS class.

How tiled windows look now (`tiled` CSS class missing and persistent rounded corners)...

![tiled-class-not-set](https://user-images.githubusercontent.com/3822556/130440572-0ae5b3f8-7ed4-4cb2-ac69-e0a5a24756f3.png) ![tiled-class-not-set-debug](https://user-images.githubusercontent.com/3822556/130440569-5b3981b5-b40a-477f-84c7-80d3833ad6b8.png)

How tiled windows used to look like (`tiled` CSS class present and squared corners)...

![tiled-class-set](https://user-images.githubusercontent.com/3822556/130440576-7f17c64c-ac3d-4313-8dde-c88f0dbfd9cd.png) ![tiled-class-set-debug](https://user-images.githubusercontent.com/3822556/130440573-70d271dd-f100-436c-91a7-fc10f89a82f7.png)

### Steps to reproduce

Using any default Linux Mint theme (more noticeable when using Mint-X), open any application using CSD windows and tile it anywhere. The border radius of headerbar corners should be removed but it isn't.

[Here is a little Python application to perform tests in case it is needed.](https://github.com/linuxmint/muffin/files/7031536/headerbar.py.gz)

### Other information

- I tested Cinnamon versions 4.6.x (LM 20), 4.8.x (LM 20.1) and 5.0.x (LM 20.2) and in all of these the `tiled` CSS class was **not** applied to tiled windows.
- I tested Cinnamon versions 3.0.x, 3.2.x and 3.6.x and in all of these the `tiled` CSS class was **correctly** applied to tiled windows.

### Extra information that most likely is irrelevant

- Under Linux Mint 20.2 XFCE using XFwm as window manager the `tiled` CSS class was **not** applied to tiled windows. Using Metacity as window manager the `tiled` CSS class was **correctly** applied.
[...]

PR #598, opened 2021-09-06
window.c: Properly update gtk_edge_constraints

JosephMcc:
We support the atom but it was never being updated. This resulted in csd
windows not having their style properly updated when tiled or maximized.


JosephMcc:
Fixes: https://github.com/linuxmint/muffin/issues/596

PR #626, opened 2022-07-26
WIP/RFT Revert renaming of the X11 atoms wm_name and gnome_wm_keybindings

Cobinja:
This reverts two X11 atoms back to how they were in 5.2
This needs some testing to see if there are any negative side effects.


mtwebster:
I'm going to merge this before we tag 5.6.. it seems fine to me

PR #638, opened 2022-09-05
Support multiple key bindings for 'locate-pointer-key'

rcalixte:
In tandem with linuxmint/cinnamon#11143

### DIFF
--- a/src/core/prefs.c
+++ b/src/core/prefs.c
@@ -1719,29 +1719,44 @@ locate_pointer_key_handler (GVariant *value,
                             gpointer *result,
                             gpointer  data)
 {
-  MetaKeyCombo combo;
-  const gchar *string_value;
+  MetaKeyCombo *combo;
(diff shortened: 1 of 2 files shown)


src/core/prefs.c
```diff
@@ -1719,29 +1719,42 @@ locate_pointer_key_handler (GVariant *value,
...
-  if (locate_pointer_key_combo.keysym != combo.keysym ||
-      locate_pointer_key_combo.keycode != combo.keycode)
+          g_free (combo);
+
+          output = FALSE;
+        }
+
+      combo->modifiers = 0;
+    }
+
```
mtwebster, line 1749: I think you need a `g_free` on string_values after you're done with it - to free the array itself (but not the values).

rcalixte: Apologies @mtwebster, my C is rusty. This should be fixed now.

PR #661, opened 2023-02-11
meta-close-dialog-default.c: Delete html bold tag

### LINKED ISSUE #660: Close dialog shows HTML code
```
 * muffin version - 5.6.3
 * Distribution - Mint 21.1
 ```

**Issue**
When is opening the window to force the application exit, this shows HTML code which does not fulfill its function of bold the text. There is a "< b >Application is not responding< / b >" but should be <b> Application is not responding </b> .

ecsuarez:
Close #660


mtwebster:
Are you running muffin with something other than Cinnamon? You should never run into this otherwise.

Also, if I force the default dialog to be used, I can't reproduce the original issue (either via Cinnamon or muffin --replace):
image

PR #681, opened 2023-12-14
[WIP] wayland: add support for idle-inhibit

Sunderland93:
Adapted from https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/3145


Sunderland93:
At the moment I can’t test it due to the lack of Wayland support in cinnamon-screensaver

mtwebster:
There's probably never going to be a cinnamon-screensaver during a Wayland session. We're currently planning on integrating it into Cinnamon directly.

I don't mind keeping this open for now but I don't really have a timeline for addressing the lack of a screensaver.

PR #696, opened 2024-06-28
constraints.c: Allow resize of windows moved above screen

jknockel:
If the titlebar of a window has been moved above the screen by a user via an unconstrained move, then any constrained user resize following this move will cause the window to jump below the top of the screen or cause other glitchy behavior.

This commit removes the constraint that the titlebar of a window must be below the top of the screen for any resize that is both (1) triggered by a user and (2) is a resize that affects only the left, right, or bottom edges of the window.  This allows users to move a window partially above the screen and then resize the window to be wider or resize the bottom edge of the window to make it taller or shorter.

This is a backport of 5ba364a9476905d9b92e7dc6cdc2f273bcda2498.

### DIFF
--- a/src/core/constraints.c
+++ b/src/core/constraints.c
@@ -1720,6 +1720,7 @@ constrain_titlebar_visible (MetaWindow         *window,
                             gboolean            check_only)
 {
   gboolean unconstrained_user_action;
+  gboolean user_nonnorthern_resize;
   gboolean retval;
   int bottom_amount;
   int horiz_amount_offscreen, vert_amount_offscreen;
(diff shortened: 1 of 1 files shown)


mtwebster:
Can you describe better how to reproduce the undesirable behavior? 

Currently I can move a window above the screen edge using alt-drag. I can then resize that window by grabbing any visible edges/corners, without affecting its position (it does not jump back into the work area).

jknockel:
Interesting!  This issue appears to only affect CSD windows (e.g., I can see it in gedit and Google Chrome as two examples).  Can you reproduce the issue with a CSD window?

PR #710, opened 2025-01-09
wayland: send key-down events for modifier keys

xmvziron:
See my post on the linked issue for details.

Closes: linuxmint/wayland#1

### DIFF
--- a/src/core/keybindings.c
+++ b/src/core/keybindings.c
@@ -2203,6 +2203,10 @@ process_special_modifier_key (MetaDisplay          *display,
                            clutter_input_device_get_device_id (event->device),
                            XISyncDevice, event->time);
 
+          /* treat modifier keys the same as normal keys on Wayland */
+          if (meta_is_wayland_compositor ())
+            return FALSE;
+
           return TRUE;
         }
       else


ezterry:
This fixed some issues I was having with wine/proton and the game not recieving modifier keys (Shift, Control, or Alt)

However it causes an unexpected behavior with the meta [windows] key: The menu opens on keydown, thus causing keybinds based on meta to be hard to trigger (it will trigger if closing the menu) This includes window movement shortcuts (meta-left/right) as well as custom shortcuts I have to launch apps)

Sunderland93:
Unfortunately, this fix breaks the existing behavior of the Super modifier key: a single Super/Win modifier should only work on release, otherwise it breaks the work of keybindings with this modifier (for example, Win + left mouse click to drag a window, or combinations to attach windows to the screen edges). However - all other single modifiers should work on press. In this section of code, a check is made whether a single modifier is pressed, and if so, a KeyRelease event is sent. The only problem is that this applies to all modifiers described in the `modifier_only_keyval function`. And this only concerns work in Wayland, in X11 processing is performed as needed. I have not yet found a way to fix this behavior for Wayland.

https://github.com/linuxmint/muffin/blob/051b8d20e8f958c85af83a79f5883446749c2709/src/core/keybindings.c#L2244

mtwebster:
This will need fixed in muffin, as this has too many side-efffects as mentioned.

Issue #701, opened 2024-08-12
Wine Fullscreen Apps are not Unredirected Correctly


li20034:
### Distribution

Mint 22

### Package version

6.2.9

### Graphics hardware in use

Intel HD 530

### Frequency

Always

### Bug description

Fullscreen apps running in Wine are still being composited likely negatively impacting performance and latency, rather than being unredirected so the app draws directly to the screen.

### Steps to reproduce

1. Turn on `Settings -> General -> Compositor Options -> Disable Compositing for full-screen Windows`
2. Run a 3D demo or game in Wine in fullscreen. Something which runs at a high framerate and vsync off is desirable to really see the effects.
3. ssh into the machine from another device
4. in the ssh session, run `sudo intel_gpu_top`. You will observe that cinnamon is listed with a significant amount of 3D Render usage. Also there is no tearing even with vsync off in the app. This means that cinnamon is still compositing the window.
5. For reference, try running the native app `glmark2 --fullscreen` instead of Wine. This time, the GPU render usage for cinnamon will go to (almost) 0 after a second or 2.

### Expected behavior

Cinnamon should use almost 0 GPU in this scenario, and the window gets unredirected, meaning that with VSync off, tearing should be visible. 

### Additional information

I'm using Wine 9.0 from the official Ubuntu repos (not the winehq ppa). I listed the GPU as Intel HD 530 (on i5 6600), but the same issue also occurs on other Intel GPUs (e.g. Intel Iris Xe on i7-1180G7). I'm almost sure that this issue occurs with other GPUs too, but I do not have the hardware to test it.

li20034:
That's definitely interesting. So it's likely not an issue with Wine/Proton itself then, since KDE works. One thing I'm curious about is whether this happens on Gnome, since cinnamon and muffin are forks of gnome-shell and mutter respectively. I haven't had the chance to test that yet since I don't use Gnome.

The other thing I'm curious about is whether this issue occurred in Linux Mint 21. Because a few years ago, there was a very similar issue to this (in terms of symptoms) described [here](https://github.com/linuxmint/muffin/issues/544), which was fixed in late 2021.

Enokilis:
> The other thing I'm curious about is whether this issue occurred in Linux Mint 21. Because a few years ago, there was a very similar issue to this (in terms of symptoms) described [here](https://github.com/linuxmint/muffin/issues/544), which was fixed in late 2021.

[...]

PR #713, opened 2025-02-09
Fix inability to unredirect OpenGL fullscreen Wine/Proton/native Steam games and ignoreance of unredirection setting.

itz-me-zappex:
`meta_surface_actor_x11_should_unredirect()` returns false to avoid unredirection for any kind of window if unredirection for fullscreen windows disabled through Cinnamon settings (or through `dconf`).

`meta_window_is_screen_sized()` obtains focused window rectangles if those are broken in currently handled child window (`x=0, y=0` and `1x1`) and then checks whether window is screen sized or not.

Partially closes [701](<https://github.com/linuxmint/muffin/issues/701>).

### DIFF
--- a/src/compositor/meta-surface-actor-x11.c
+++ b/src/compositor/meta-surface-actor-x11.c
@@ -254,6 +254,9 @@ meta_surface_actor_x11_is_opaque (MetaSurfaceActor *actor)
 gboolean
 meta_surface_actor_x11_should_unredirect (MetaSurfaceActorX11 *self)
 {
+  if (!meta_prefs_get_unredirect_fullscreen_windows ())
+    return FALSE;
+
   MetaWindow *window = self->window;
 
   if (meta_window_requested_dont_bypass_compositor (window))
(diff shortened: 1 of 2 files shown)


itz-me-zappex:
Turns out that unredirection is impossible for fullscreen Wine/Proton/native Steam games __using OpenGL__ (vanilla muffin 6.4.1), patch for `src/core/window.c` allows BLIT unredirection for those OpenGL games, but the most interesting thing is that muffin already unredirects Vulkan apps perfectly (meaning allows FLIP), except unredirection will work even if that disabled in settings (check `src/compositor/meta-surface-actor-x11.c` patch).

And `xwininfo -tree -root` shows that Vulkan games are having proper rectangles on child windows, unlike OpenGL ones with `1x1+0+0`, weird.

So I could assume that something else also behaves weird on muffin side, since if unredirection allowed, it should use FLIP for OpenGL too, but that very hard to figure with that out.

itz-me-zappex:
Forget that, I found better solution which allows FLIP for such windows, and does not require that mess, I will create new PR, sorry for spamming PRs...

Issue #656, opened 2022-11-29
Applications fullscreen to wrong monitor when window maximized through tiling


isRedacted:
* Cinnamon version 5.4.12
 * Distribution - Mint 21
 * RTX 3060, NVIDIA driver version 520 but issue persists on 515 too
 * 64 bit

### Issue
When maximizing an application by dragging window to top edge of monitor and with relevant option enabled in settings, if the application is then made fullscreen it will work once and any subsequent fullscreens move to wrong monitor

### Steps to reproduce
1. Set "Maximize, instead of tile, when dragging a window to the top edge" setting on in Window Tiling
2. Open application (Brave browser, VLC, etc)
3. Drag to top
4. Make fullscreen, un-fullscreen, and fullscreen again

### Expected behaviour
Program fullscreens to monitor it is maximized to every time

### Other information
Universal to all applications that can be maximized in this way, issue only present on NVIDIA drivers (tested on xserver version 1:1.0.17-2 and works correctly)

pilot51:
This used to not be a problem for me and it started happening after upgrading to Mint 21. I watch a lot of YouTube and that's where I experience it most if not all of the time. I usually have Firefox maximized on my right/secondary monitor and the video goes fullscreen to my left/primary monitor. I work around it by unmaximizing and maximizing the window and then it will fullscreen on the right monitor.

VividLemon:
It would be nice if this issue got some more attention. It is quite annoying of an issue, that occurs, to the very least, once every time I use my machine

mistrjirka:
I also get this issue.

lizardlab:
I have similarly gotten this issue on my multi-monitor setup with **Firefox** specifically on any website that full screens, could be Youtube, Tubi, Vimeo etc.

AndrewRutherfoord:
I'm having this issue in Chrome and noticed it mainly in YouTube when full screening a video, but the same thing happens when I click F11 on a chrome window. I have to do the same thing as @pilot51 to get it to maximize on my second monitor.

ScottDemarest:
I'm getting this issue with fullscreening with okular pdf reader. Also a similar issue with gaomon drawing tablet driver not being able to select particular screens. A fix would be nice

AndreiTS:
Having this issue on firefox, going fullscreen on any site. I have to restart cinammon to work again.

ColdSpirit0:
[...]

PR #714, opened 2025-02-09
window.c: Update saved_rect to target monitor work area on tiling

CollinBartlam:
This should fix an issue (#656) where in some scenarios after tiling across monitors, windows would full screen onto the wrong monitor. The change updates the meta_window_tile so that saved_rect and unconstrained_rect are adjusted based on the target monitor’s work area which ensures consistent behaviour on the intended monitor. While this approach resolves the issue, I haven't worked with Muffin's codebase before, so feedback on improving the implementation is welcome.

### DIFF
--- a/src/core/window.c
+++ b/src/core/window.c
@@ -3541,21 +3541,43 @@ meta_window_tile (MetaWindow   *window,
     directions = META_MAXIMIZE_VERTICAL;
 
 
-  GdkRectangle *maybe_saved_rect = NULL;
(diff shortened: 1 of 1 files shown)


mtwebster:
I'm not sure what's going on with this, but it ends up restricting the usable space on one of my monitors to the lower half of it only.  If I disable all but one monitor, it restricts that also (even though the wallpaper, panels, etc.. are fullscreen).

Issue #607, opened 2021-10-20
Regression- Games that are running way above of 60 fps hard locks the system


Leopard1907:
* muffin version :muffin 5.0.2
 * Distribution: Mint 20.2
 * Graphics hardware *and* driver used: https://gist.github.com/Leopard1907/c3d8ef9451cf5c82723f7f0533f40412
 * 32 or 64 bit: 64 bit OS
 * System info: https://termbin.com/gf9p

**Issue**
After latest Muffin update dropped at 2021-10-16 , games that are running way above of 60 fps locks up entire system. Moreover, disabling/enabling compositing setting at "General" settings of OS doesn't seem to have any effect. 

**Steps to reproduce**

I can consistently reproduce the issue with Half Life 2 from Steam. Running Steam client with ```__NV_PRIME_RENDER_OFFLOAD=1 __GLX_VENDOR_LIBRARY_NAME=nvidia __VK_LAYER_NV_optimus=NVIDIA_only``` to make sure everything runs on Nvidia dgpu. This is a prime system, at on-demand mode.

**Expected behaviour**

Not crashing,freezing,hanging

**Other information**

Game is set to fullscreen. Vsync is off in game, vsync is on on Cinnamon DE settings. Changing Vsync methods doesn't seem to have any effect.

Leopard1907:
I've tested with KDE Plasma on same system, using Kubuntu's KDE backports ppa. Everything runs fine there. So it is a Cinnamon/Muffin issue.

clefebvre:
Can you confirm that going back to muffin 5.0.1 fixes the issue?

mtwebster:
My system seems to lock up on  HL2 *no matter which version of muffin I'm using*

- installed HL 2
- running on-demand NVIDIA GM107GLM [Quadro M1200 Mobile] with 470.74 driver
- run: `__NV_PRIME_RENDER_OFFLOAD=1 __GLX_VENDOR_LIBRARY_NAME=nvidia __VK_LAYER_NV_optimus=NVIDIA_only steam steam://rungameid/220`
- In nvidia settings I'm using 'adaptive' power mode.

![image](https://user-images.githubusercontent.com/262776/139733991-9472e178-f4b8-4517-a522-512ceece0c03.png)

It runs anywhere from not even getting past the initial new/load game dialog  to a minute or so in game. The freezes are complete system locks, though the mouse still responds. I need to cycle power when it happens.

Is this the same behavior you get in 5.0.2?

[...]

PR #715, opened 2025-02-11
[WIP] [X11] Workaround 1x1 Wine window to allow unredirect.

### LINKED ISSUE #701: Wine Fullscreen Apps are not Unredirected Correctly

itz-me-zappex:
#### Description
Finally, after previous two PRs I have created (and closed) with chunky workarounds and wild activity in #701, I have found solution and figured out that root of issue is bugged rectangles of child windows (`1x1+0+0`), which are created by OpenGL games running through Wine/Proton (e.g. Geometry Dash) and even native ones running through Steam (e.g. native Terraria). That does not happen with OpenGL games running completely natively (i.e. w/o Steam container) and does not happen with Vulkan games at all as those are have proper rectangles on child windows as `xwininfo -tree -root` shows, so solution oriented only at that specific case.

#### Solution
So, the best way I found to solve this, is just not override currently handled window rectangles with new (child) ones if those are bugged (i.e. resolution of child reported as `1x1`), that makes muffin use parent window rectangles which are correct.

That "filter" should be used two times:
1. in `window.c`: to allow FLIP unredirection for such windows by handling those with proper rectangles. 
2. in `window-x11.c`: to avoid unredirection breakage if window mode appears changed e.g. from fullscreen to windowed and back, or when OSD stuff draws over unredirected window.

Also I fixed issue in `meta-surface-actor-x11.c` related to handling internal unredirect requests if "Disable compositing for full-screen windows" disabled in settings (or through dconf), as those requests always have been handled and fullscreen windows were unredirected.

Closes #701.

### DIFF
--- a/src/core/window.c
+++ b/src/core/window.c
@@ -1100,25 +1100,30 @@ _meta_window_shared_new (MetaDisplay         *display,
   /* avoid tons of stack updates */
   meta_stack_freeze (window->display->stack);
 
-  window->rect.x = attrs->x;
-  window->rect.y = attrs->y;
-  window->rect.width = attrs->width;
-  window->rect.height = attrs->height;
+  if (attrs->width != 1 &&
+      attrs->height != 1)
+    {
+      window->rect.x = attrs->x;
+      window->rect.y = attrs->y;
+      window->rect.width = attrs->width;
+      window->rect.height = attrs->height;
 
-  /* size_hints are the "request" */
-  window->size_hints.x = attrs->x;
-  window->size_hints.y = attrs->y;
-  window->size_hints.width = attrs->width;
-  window->size_hints.height = attrs->height;
-  /* initialize the remaining size_hints as if size_hints.flags were zero */
-  meta_set_normal_hints (window, NULL);
+      /* size_hints are the "request" */
+      window->size_hints.x = attrs->x;
+      window->size_hints.y = attrs->y;
+      window->size_hints.width = attrs->width;
+      window->size_hints.height = attrs->height;
 
-  /* And this is our unmaximized size */
-  window->saved_rect = window->rect;
-  window->unconstrained_rect = window->rect;
+      /* initialize the remaining size_hints as if size_hints.flags were zero */
+      meta_set_normal_hints (window, NULL);
 
-  window->depth = attrs->depth;
-  window->xvisual = attrs->visual;
+      /* And this is our unmaximized size */
+      window->saved_rect = window->rect;
+      window->unconstrained_rect = window->rect;
+
+      window->depth = attrs->depth;
+      window->xvisual = attrs->visual;
+    }
 
   window->title = NULL;
   window->icon = NULL;
--- a/src/x11/window-x11.c
+++ b/src/x11/window-x11.c
@@ -3817,10 +3817,14 @@ meta_window_x11_configure_notify (MetaWindow      *window,
   g_assert (window->override_redirect);
   g_assert (window->frame == NULL);
 
-  window->rect.x = event->x;
-  window->rect.y = event->y;
-  window->rect.width = event->width;
-  window->rect.height = event->height;
+  if (event->width != 1 &&
+      event->height != 1)
+    {
+      window->rect.x = event->x;
+      window->rect.y = event->y;
+      window->rect.width = event->width;
+      window->rect.height = event->height;
+    }
 
   priv->client_rect = window->rect;
   window->buffer_rect = window->rect;


Enokilis:
Would it also be possible to fix the root of the issue?
Why does Proton/Wine create a 1x1 window in the first place? I can actually see it in the corner, manifesting as the desktop's background color.

itz-me-zappex:
Just summarized commit.

itz-me-zappex:
As far as I understand, fixing it not planned at all, right? To be more correct - workarounding. And #701 issue is still opened for around 9 months and successfully ignored by development team.

I have found report on Wine's Bugzilla created in 2021
https://bugs.winehq.org/show_bug.cgi?id=50717

And my changes aimed to ignore this 1x1 window, otherwise window will not be unredirected, so that compositor will reduce performance and cause microstuttering on monitor side.

I can confirm that is weird that native games from Steam also have the same issue. And again, I will add that issue is not reproducible with games using Vulkan API (including those running with DXVK or VKD3D), only OpenGL stuff suffers from that.

In case someone uses Arch Linux like me, you can build and install my fork:
```bash
# Official PKGBUILD from Arch Linux GitLab
git clone https://gitlab.archlinux.org/archlinux/packaging/packages/muffin

# Go to directory with PKGBUILD
cd muffin/

# Install build dependencies, download and extract source code 
makepkg --syncdeps --nobuild

# Replace extracted source code with one from my fork
rm -rf src/muffin-6.4.1
git clone https://github.com/itz-me-zappex/muffin-fixed-unredirection.git src/muffin-6.4.1

# Build and install
makepkg -i
```

Or just modify PKGBUILD by skipping checksums and replacing source with one from my repository.

After installation press `Ctrl+Alt+Esc` to apply changes.

Enokilis:
Followed the steps, but I still see a visible 1x1 window in the corner and get no flipping.

voron00:
+1 for this, without this patch gsync isn't working in games.

itz-me-zappex:
Okay, no problem, I will revert changes in `src/compositor/meta-surface-actor-x11.c`.

mtwebster:
Hi I did some testing on this. For now I reverted the repositioning of that settings check so I could see just how the 1x1 workarounds would affect things.

I can see it had an impact with Geometry Dash, but every other game I've tested (so far) unredirects fine at fullscreen on git master, or wouldn't unredirect, and neither version made a difference.

I also observed the check in `meta_window_x11_configure_notify()` running into unrelated windows, which was puzzling (but didn't seem to cause an issue).

I'm sort of leaning into trying to make it a bit easier to force a window to un-redirect, as it seems as though there will always be exceptions. I worry about side effects of these small workarounds with other applications (who may also be doing odd things with windows).

itz-me-zappex:
>Hi I did some testing on this. For now I reverted the repositioning of that settings check so I could see just how the 1x1 workarounds would affect things.

Hi, I already reverted changes related to handling unredirect and kept only workaround.

> I can see it had an impact with Geometry Dash, but every other game I've tested (so far) unredirects fine at fullscreen on git master, or wouldn't unredirect, and neither version made a difference.

Yes, especially noticeable in Geometry Dash, this is why I started figuring with this out, and as I said in PR's description, issue affects all OpenGL games running through Wine/Proton (Lutris, Steam, Heroic etc.), and even native ones that running from Steam (those are FNA by the way). No idea why to be honest. Vulkan/DXVK/VKD3D games are not affected, WINED3D also not affected, despite a fact this is OpenGL.

Affected OpenGL games in my testing:
- Wolfenstein: The New Order (Proton)
- Geometry Dash (Proton)
- Terraria (native version from Steam, FNA issue?)
- Starbound (same as with Terraria)

Other games in my libraries using DXVK/VKD3D, so nothing to test anymore.

And I did not check master branch for now, still using outdated source existing in forked repo, will try update on weekends.

> I also observed the check in meta_window_x11_configure_notify() running into unrelated windows, which was puzzling (but didn't seem to cause an issue).

`¯\_(ツ)_/¯`

> I'm sort of leaning into trying to make it a bit easier to force a window to un-redirect, as it seems as though there will always be exceptions. I worry about side effects of these small workarounds with other applications (who may also be doing odd things with windows).

When I tinkered with source code (two previous PRs were shitty so I closed them), I noticed that I could not obtain proper window rectangles "manually" using already existing functions (i.e. explicitly, not using ones from event) of parent window when 1x1 window appears. This is why I tried to get those directly from X11.

I still did not notice any side effect of this workaround (despite a fact `xwininfo -tree -root` shows a lot of `1x1` windows), but yeah, needs more testing, I remember I have seen one PR here which fixed [similar issue](https://github.com/linuxmint/muffin/issues/544) and caused [system lock up](https://github.com/linuxmint/muffin/issues/607) on some systems.

Here is one guy who did not report any issue on this workaround and said it fixed not working GSYNC (makes a sense, because window should be unredirected for this), and this comment has 2 thumbs up (mine does not count). [Here](https://github.com/linuxmint/muffin/pull/715#issuecomment-2970743637).

itz-me-zappex:
Probably we need to create a function that checks for whether window has `1x1px` rectangles or not, so in case something it will be easy to find this workaround in the code, also because these checks look ugly.

mtwebster:
Addressed in 1b2362dee0366e9 and further in cc8baadc18db6e.

PR #719, opened 2025-04-11
Revert "x11 backend: Don't let muffin reset the keyboard layout..."

mtwebster:
This reverts commit e6bcb95cc2b005223c60e8b3157f0e9f1fb56d86.

Layouts will no longer be handled by libgnomekbd.

ref: https://github.com/linuxmint/cinnamon/pull/12758


pewgar:
other than keyboard layouts, wayland support seems pretty usable and good

PR #725, opened 2025-05-16
[Wayland] Implement xdg-activation-v1 protocol

Sunderland93:
Adapted from Mutter with some additional changes https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1845


Sunderland93:
This part works well. If you start Firefox on the first workspace, switch to the second workspace and run, for example, `System Info` there, where you click on `Upload system information` - the Firefox window on the first workspace receives an activation token (according to the output of `WAYLAND_DEBUG=1`), and its indicator on the panel starts to glow (urgent state). However, there is no automatic switch to the first workspace and focus on the window with Firefox, it still requires manual switching to the first workspace and maximizing the window. I think this is a problem of Cinnamon, which cannot yet correctly handle window activation. @mtwebster any suggestions?

mtwebster:
It looks like in your test case the activation is being caught in [windowAttentionHandler.js](https://github.com/linuxmint/cinnamon/blob/master/js/ui/windowAttentionHandler.js), and `meta_window_activate()` is called. The timestamp is 0 so I think it ends up returning at line 4174 in window.c. 

In an x11 session, the window does *not* go thru windowAttentionHandler, but activation occurs from somewhere else, with a valid timestamp, so workspace switching can occur. I haven't had a chance to figure out what's causing that activation yet, and maybe it shouldn't be the same one wayland uses - getting a proper timestamp may be what we need to figure out.

mtwebster:
Sorry can you rebase this once more, please? I'd like to get it merged even though it's not quite 1:1 with x11. Thanks

Issue #669, opened 2023-07-03
After the rebase onto `mutter 3.36.0-0ubuntu0.20.04.1`, there are mentions of mutter remaining in the CLI interface


audibleptr:
After the https://github.com/linuxmint/muffin/commit/8599127fa99dc285426fd84c2f53ed0ae8d60d3c rebase, there are now at least a few mentions of _mutter_ remaining in the CLI interface (see e.g. the https://github.com/linuxmint/muffin/blame/94a950abd1f853d1f0d0ca5d2133c58befc6b363/src/core/mutter.c#L38 file).

For instance, `mutter 5.8.0` in:

```
$ muffin --version
mutter 5.8.0
Copyright © 2001-2011 Havoc Pennington, Red Hat, Inc., and others
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

And, more importantly, the CLI argument `--mutter-plugin=PLUGIN     Mutter plugin to use` also changed its name in:

```
$ muffin --help
Usage:
  muffin [OPTION?]

Help Options:
  -h, --help                 Show help options

Application Options:
  --sm-disable               Disable connection to session manager
  -r, --replace              Replace the running window manager
  --sm-client-id=ID          Specify session management ID
  -d, --display=DISPLAY      X Display to use
  --sm-save-file=FILE        Initialize session from savefile
  --sync                     Make X calls synchronous
  --x11                      Run with X11 backend
  --version                  Print version
  --mutter-plugin=PLUGIN     Mutter plugin to use
```

This doesn't cause any issue in usage _per se_, but not sure whether this was an intended change in muffin 5.4.0 (persisting until the current version in `master` @ https://github.com/linuxmint/muffin/commit/94a950abd1f853d1f0d0ca5d2133c58befc6b363).

PR #726, opened 2025-05-16
Replace any references to mutter with muffin

Sunderland93:
Fixes https://github.com/linuxmint/muffin/issues/669

I'll test it more this weekend. But running a nested `muffin` works without problems.


leigh123linux:
-1 to doing a rename.
I removed the muffin binary in fedora packaging as it's use is unsupported.

@mtwebster  Isn't it time to make muffin libs only?

Sunderland93:
> I removed the muffin binary in fedora packaging as it's use is unsupported.

`Muffin` e.g. can be launched as a nested Wayland compositor, for testing purposes. Just like `Mutter`. It'll be very useful while implementing various Wayland protocols in Muffin.

PR #727, opened 2025-05-17
[Wayland] Bump wl_output interface to v4

Sunderland93:
wl_output v3 allows clients to tell the
server that it can clean up the related wl_resource.

wl_output v4 adds the name and description events already present in xdg_output. Some clients may rely on that information instead of xdg-output-v1


Sunderland93:
Tested, works well, no regressions found

Screenshot from 2025-05-17 13-13-32

PR #740, opened 2025-08-13
[Wayland] Add support for xdg_toplevel.configure_bounds()

Sunderland93:
This PR adds support for `xdg_toplevel_configure_bounds()`, which is aims to communicate the maximum size a surface should be created with, and loosely corresponds to the concept of "work area" in the X11 world. It was [introduced](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/41) in v4 of xdg-shell protocol, so also bump it's version.

Adapted from Mutter: https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/2167


Sunderland93:
Tested on my machine, works pretty well. I'll plan to add unit tests for it (and other features) in a separate PR's, since it requires additional changes.

PR #744, opened 2025-08-20
[Wayland] Direct scanout

Sunderland93:
This is my second attempt to backport support for unredirecting fullscreen Wayland surfaces from GNOME. All issues I encounter with previous PR is solved, Cinnamon runs fine. Although it still need some tests. I split this PR in separate commits to facilitate reviewing and keep original commit messages from GNOME, as well as links to the original Mutter MR's.

https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/798

Also I'm not sure if this [part ](https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/721)is also needed to be backported to Cinnamon

Please note that this PR supports only legacy modesetting API (atomic will be added later)

### DIFF
--- a/src/compositor/meta-compositor-native.h
+++ b/src/compositor/meta-compositor-native.h
@@ -25,7 +25,7 @@
 
 #define META_TYPE_COMPOSITOR_NATIVE (meta_compositor_native_get_type ())
 G_DECLARE_FINAL_TYPE (MetaCompositorNative, meta_compositor_native,
-                      META, COMPOSITOR_NATIVE, MetaCompositor)
+                      META, COMPOSITOR_NATIVE, MetaCompositorServer)
 
 MetaCompositorNative * meta_compositor_native_new (MetaDisplay *display);
 


Sunderland93:
> Hm, with latest `master` session is crashed at startup

Nevermind, fixed now. Ready for review and test

leigh123linux:
@mtwebster Does it crash if you disable eglstream support in muffin, IMO no one should use it.

Sunderland93:
@leigh123linux you also need to build `Cinnamon` and `cinnamon-desktop` from `master`, as this PR contains some API changes which is merged in `master`

leigh123linux:
@Sunderland93  The only difference in cinnamon-desktop is this commit, my installation includes it.

https://github.com/linuxmint/cinnamon-desktop/commit/879412ae7299ff43d187a7f461a11183c4376d5c

Sunderland93:
I can't reproduce this crash on QEMU and my desktop with Radeon RX 570, in both Xorg and Wayland sessions. Will test on my laptop with Intel later.

src/backends/native/meta-renderer-native.c
```diff
@@ -2241,6 +2270,92 @@ meta_renderer_native_create_dma_buf (CoglRenderer  *cogl_renderer,
...
+  g_warn_if_fail (!onscreen_native->gbm.next_fb);
+  g_set_object (&onscreen_native->gbm.next_fb, META_DRM_BUFFER (scanout));
+
+  ensure_crtc_modes (onscreen, kms_update);
+
+  onscreen_native->pending_queue_swap_notify_frame_count =
+    renderer_native->frame_counter;
+  meta_onscreen_native_flip_crtcs (onscreen, kms_update);
+
+  meta_kms_post_pending_update_sync (kms);
```
mtwebster: Should this call post_pending_update(kms) instead?

mtwebster:
Seems ok now. I can observe this working while playing a fullscreen video with PAINT_DAMAGE_REGION debugging enabled.

I wonder if there's any justification to hooking it up to Cinnamon's 'disable compositing for fullscreen windows'... or that should just become an x11 relic.  This actually is doing a better job - it applies to fullscreen celluloid videos, where as unredirection does not apply in x11. That could be some issue with fullscreen detection on that window, since it's embedding mpv.

PR #758, opened 2025-10-21
Update builddeps: Use libglvnd instead of mesa deps

leigh123linux:


Fantu:
I did now a fast search, libgl1-mesa-dev and libgles2-mesa-dev are transition packages that have deps libgles-dev and libglvnd-dev, and libglvnd-dev have libgles-dev as dep, so your change doesn't seem to introduce any "real changes" to the graphic stack packages and it should be fine.

PR #773, opened 2025-12-19
Wayland: Fix nvidia refresh rate when running nvidia driver

### LINKED ISSUE #130: 30 fps
### Distribution

Linux Mint 22.1 Cinnamon

### Package version

Cinnamon 6.4.8

### Graphics hardware in use

nvidia 4060

Driver: 570.86.16

```
sudo add-apt-repository ppa:graphics-drivers/ppa
sudo apt update
```

### Frequency

Always

### Bug description

In Wayland session, the screen refreshes at 30 fps instead of 60. The screen settings are set to 60 fps.

The cursor moves jerkily, glxgears shows 30 fps.

p.s. Steam crash desktop on startup.

leigh123linux:
Fixes https://github.com/linuxmint/wayland/issues/130

This switches the order of what renderer mode is tried first, so that
the gbm renderer mode is preferred on an NVIDIA driver where it is
supported.
We fall back to still try the EGLDevice renderer mode if the created gbm
renderer is not hardware accelerated.
The last fallback is still to use the gbm renderer, even if it is not
hardware accelerated, as this is needed when hardware acceleration isn't
available at all. The original reason for the old order was due to the
fact that a gbm renderer without hardware acceleration would succeed
even on NVIDIA driver that didn't support gbm.

Based on this commit.

https://gitlab.gnome.org/GNOME/mutter/-/commit/fdfd2c6bfe553d35a69e07599be597c95e28d58c

### DIFF
--- a/src/backends/native/meta-renderer-native.c
+++ b/src/backends/native/meta-renderer-native.c
@@ -3698,29 +3698,29 @@ meta_renderer_native_create_renderer_gpu_data (MetaRendererNative  *renderer_nat
   GError *egl_device_error = NULL;
 #endif
 
-#ifdef HAVE_EGL_DEVICE
-  /* Try to initialize the EGLDevice backend first. Whenever we use a
-   * non-NVIDIA GPU, the EGLDevice enumeration function won't find a match, and
-   * we'll fall back to GBM (which will always succeed as it has a software
-   * rendering fallback)
+  /* Try to initialize the GBM backend first. GBM provides better compatibility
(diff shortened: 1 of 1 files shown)


Milli189:
any update from this?

Mosnar23:
"All checks have passed"
"No conflicts with base branch"
"1 approval"
;-;
(It works too, I'm testing it rn)

Mosnar23:
Awesome :)

PR #796, opened 2026-02-09
wayland/subsurface: Move actor unparenting back to rebuild_surface_tr…

Sunderland93:
…ee()

Unparenting the surface actor when the subsurface object is destroyed has several issues:

subsurface actors can be unparented while a close animation is still ongoing, breaking the animation for e.g. Firefox. adding and removing the actor to/from the parent is not handled in one place, making the code harder to follow.
if the destroyed subsurface had children of its own, they potentially stick around until a surface-tree rebuild. This makes the Firefox hamburger menu not close with the "compositor" backend.

Move the unparenting back to
`meta_window_actor_wayland_rebuild_surface_tree()` and instead just notify the parent of a state change, if it still exist. This will ensure a correct mapping between the subsurface node tree and the flat surface actor list. In case of the closing animation the parent will already be removed and the call is skipped.

### DIFF
--- a/src/compositor/meta-window-actor-wayland.c
+++ b/src/compositor/meta-window-actor-wayland.c
@@ -39,28 +39,43 @@ typedef struct _SurfaceTreeTraverseData
 } SurfaceTreeTraverseData;
 
 static gboolean
+get_surface_actor_list (GNode    *node,
+                        gpointer  data)
+{
+  MetaWaylandSurface *surface = node->data;
+  MetaSurfaceActor *surface_actor = meta_wayland_surface_get_actor (surface);
+  GList **surface_actors = data;
+
+  *surface_actors = g_list_prepend (*surface_actors, surface_actor);
+  return FALSE;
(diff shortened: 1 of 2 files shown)


src/compositor/meta-window-actor-wayland.c
```diff
@@ -39,28 +39,43 @@ typedef struct _SurfaceTreeTraverseData
 } SurfaceTreeTraverseData;
 
 static gboolean
-set_surface_actor_index (GNode    *node,
-                         gpointer  data)
+get_surface_actor_list (GNode    *node,
+                        gpointer  data)
 {
   MetaWaylandSurface *surface = node->data;
   MetaSurfaceActor *surface_actor = meta_wayland_surface_get_actor (surface);
```
mtwebster: This can return NULL, is that ok?

I guess maybe just return FALSE in that case...

Sunderland93:
Done

PR #820, opened 2026-04-28
Fix #819: preserve user-applied monitor config across DRM hotplug

### LINKED ISSUE #819: Bug: Muffin aggressively overrides user-disabled display state upon DRM hotplug events (Monitor Input Polling)
### Distribution

Linux Mint 22.3 - Cinnamon 64-bit

### Package version

Cinnamon 6.6.7

### Graphics hardware in use

Acer Aspire A515-47 V1.04 0000000000000000 - Advanced Micro Devices, Inc. [AMD/ATI] Barcelo

### Frequency

Always

### Bug description

Muffin fails to persist a user-defined "disabled" state for external monitors. When an explicitly disabled monitor triggers a hardware hotplug event (such as during routine EDID polling/input scanning by the monitor), Muffin intercepts the event and automatically re-enables the display, overriding the user's configuration.

### Steps to reproduce

Connect a multi-input monitor (e.g., HP Omen 27q) to the laptop.

Connect a second active device (e.g., an Xbox) to the monitor's second input.

Explicitly disable the external monitor on the Linux Mint machine:

On X11: Using xrandr --output HDMI-A-0 --off or via the Cinnamon Display GUI.

On Wayland: Via the Cinnamon Display GUI.

Or just Super+P

Wait for the monitor to perform an auto-input scan (which triggers a kernel DRM hotplug event / link-state change on the PC).

### Expected behavior

Muffin should recognize that the specific hardware ID of the monitor was manually disabled by the user. It should ignore the subsequent hotplug/link-state events from that specific port and leave the display disabled.

### Additional information

Actual Behavior:
The moment the monitor scans its inputs, the kernel registers a hotplug event. Muffin immediately reacts to this event by re-applying an active configuration (e.g., re-extending the desktop and reassigning primary status).

On X11, xrandr --verbose shows the mode being re-applied seconds after it was explicitly turned off.

This effectively makes it impossible to keep the monitor turned off via software without writing custom udev rules to intercept the hardware ping before Muffin can see it.

Additional Information:
This is not a hardware defect or a kernel bug; the AMD driver is correctly reporting the state change caused by the monitor's auto-detection feature.

The issue is a policy gap in Muffin: it prioritizes the immediate hardware connection state over the user's explicit layout configuration.

This occurs regardless of whether ~/.config/cinnamon-monitors.xml is locked (read-only) or deleted, indicating the override logic is hardcoded into Muffin's hotplug response.

RedDawe:
Fixes #819.

## Background

When a user explicitly disables an external monitor (via `xrandr --output X --off`, Super+P, or the Cinnamon Display GUI), Muffin re-enables it on the next DRM hotplug event — most commonly triggered by the monitor itself performing an auto input-scan and re-toggling its link state. The user's intent is silently overridden until they re-disable, often within seconds.

## Root cause

The user's most recent intent lives in `MetaMonitorConfigManager::current_config`, which is updated on every successful `apply_monitors_config()` (including `METHOD_TEMPORARY` applies, which xrandr / Super+P / D-Bus go through). On hotplug, `meta_monitor_manager_on_hotplug()` calls `meta_monitor_manager_ensure_configured()`, whose fallback chain is `stored → suggested → previous → linear → fallback`. The currently-applied config is never consulted, so any state that hasn't also been written to `cinnamon-monitors.xml` is lost — `create_suggested()` rebuilds a fresh layout that re-enables every detected output.

This affects the common case where the user applies a temporary off (no GUI save), but also the `should_use_stored_config()` branch (`hotplug_mode_update`) where stored config is bypassed entirely.

## Fix

`src/backends/meta-monitor-manager.c`: insert a `current_config`-preferring branch at the top of `ensure_configured()`. If a config has already been applied this session and is still complete for the detected monitor set (`is_config_complete` checks the EDID-derived monitor-spec key), re-apply it. Skipped during `in_init` so the on-disk stored config still wins on session start. Genuine plug/unplug changes the spec key, so `is_config_complete` returns false and the existing fallback chain runs as before.

`src/backends/meta-monitor-config-manager.c`: guard `set_current()` against the no-op `current == config` self-set. Without this, every hotplug pushes a duplicate of the current config onto the 3-slot history queue, evicting the genuine previous configs that exist for things like `pop_previous` undo.

## Test

`src/tests/monitor-unit-tests.c`: regression test that (1) builds a config with the second monitor disabled, (2) marks it current as `apply_monitors_config()` would after a TEMPORARY apply, (3) triggers a hotplug with the same monitor topology. Asserts the disabled state is preserved across the hotplug. Without the fix this test fails because `ensure_configured()` falls through to `create_suggested()`.

Confirmation on real hardware (HP Omen 27q via amdgpu, or any monitor with auto input-scan behavior) would be appreciated from anyone with the matching setup.

### DIFF
--- a/src/backends/meta-monitor-config-manager.c
+++ b/src/backends/meta-monitor-config-manager.c
@@ -1447,6 +1447,9 @@ void
 meta_monitor_config_manager_set_current (MetaMonitorConfigManager *config_manager,
                                          MetaMonitorsConfig       *config)
 {
+  if (config_manager->current_config == config)
+    return;
+
   if (config_manager->current_config)
     {
       g_queue_push_head (&config_manager->config_history,
--- a/src/backends/meta-monitor-manager.c
+++ b/src/backends/meta-monitor-manager.c
@@ -696,6 +696,39 @@ meta_monitor_manager_ensure_configured (MetaMonitorManager *manager)
   else
     method = META_MONITORS_CONFIG_METHOD_TEMPORARY;
 
+  /* If a configuration has already been applied this session and is still
+   * valid for the detected monitor set, re-apply it. This preserves user
+   * intent across hotplug events that do not actually change the connected
+   * monitors — e.g. spurious link-state pings caused by a monitor's input
+   * auto-scan, or short DP/HDMI re-trains. Without this, the fallbacks
+   * below would compute a fresh "suggested" layout and silently re-enable
+   * a monitor the user has explicitly turned off via xrandr, Super+P, or
+   * the Display GUI. The stored on-disk configuration is consulted only
+   * during initial setup (current_config is NULL until the first apply).
+   */
+  if (!manager->in_init)
+    {
+      MetaMonitorsConfig *current_config =
+        meta_monitor_config_manager_get_current (manager->config_manager);
+
+      if (current_config &&
+          meta_monitor_manager_is_config_complete (manager, current_config))
+        {
+          if (meta_monitor_manager_apply_monitors_config (manager,
+                                                          current_config,
+                                                          method,
+                                                          &error))
+            {
+              config = g_object_ref (current_config);
+              goto done;
+            }
+
+          g_warning ("Failed to re-apply current monitor configuration: %s",
+                     error->message);
+          g_clear_error (&error);
+        }
+    }
+
   if (use_stored_config)
     {
       g_autoptr(MetaMonitorsConfig) new_config = NULL;
--- a/src/tests/monitor-unit-tests.c
+++ b/src/tests/monitor-unit-tests.c
@@ -6297,6 +6297,84 @@ meta_test_monitor_migrated_wiggle (void)
     g_error ("Failed to remove test data output file: %s", error->message);
 }
 
+/*
+ * Regression test for issue #819: a hotplug event on an unchanged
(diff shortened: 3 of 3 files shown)


RedDawe:
While looking into this I noticed muffin's `ensure_configured()` has diverged from upstream mutter, and that divergence is the actual root cause of #819.

Mutter consults the **current** (live, not-yet-persisted) config before falling back, trying configs in this order:

```
stored → [current, previous] → suggested → linear → fallback
```

Muffin currently does:

```
stored → suggested → previous → linear → fallback
```

So muffin never re-applies `current_config`, and it reaches `create_suggested` (which re-enables every output) before it ever looks at `previous`. That's why upstream isn't affected.

There are two ways to fix this, happy to go with whichever you prefer:

1. **Minimal (what this PR does now):** an early check at the top of `ensure_configured()` that re-applies the current config when the monitor topology is unchanged, before the existing fallback chain runs.

2. **Align with mutter:** rework the fallback to build an `existing_configs` list from current + previous and try those ahead of `create_suggested`, matching upstream's structure. Bigger diff and it reorders the fallback chain, but it reduces divergence and should ease future rebases.

The regression test applies either way. I have option 2 prepared on a branch and can push it as an alternative commit for comparison if that's useful.

PR #828, opened 2026-06-16
Restore edge-resistance-window setting

### LINKED ISSUE #657: Edge resistance window no longer available
```
 * muffin version: 5.6.2
 * Distribution - Manjaro Cinnamon
 ```

In the newer versions of Cinnamon I couldn't get smooth dragging windows because they were always tiling to other windows and screen edges. I have checked previous ISOs with older Cinnamon versions and it appears that the option that I was looking for is "Edge resistance with other windows and screen and monitor boundaries" and in dconf editor "edge-resistance-window" respectively.

**Other information**

I posted this issue on reddit and one of the users mentioned to reference these commits:
https://github.com/linuxmint/cinnamon/commit/bef2930e7c82587e81d13daa160a0e00bf6151c9
https://github.com/linuxmint/muffin/commit/8599127fa99dc285426fd84c2f53ed0ae8d60d3c

marcelmushik:
## What this does

Reintroduces the `edge-resistance-window` gsettings/dconf key, which was
removed in an earlier refactor. When set to `false`, the
window-to-window edge resistance threshold is set to `0`, allowing
windows to be dragged freely without the sticky-edge effect.

Defaults to `true`, so existing behavior is unchanged for anyone who
doesn't touch the setting — this is purely additive.

## Why

A number of users have asked for this toggle to come back (or to be
configurable at all). The old `gsettings set org.cinnamon.muffin
edge-resistance-window false` no longer had any effect once the key was
dropped.

Fixes #657

## Changes

- `data/org.cinnamon.muffin.gschema.xml.in` — add the `edge-resistance-window` key (default `true`)
- `src/meta/prefs.h` — add `META_PREF_EDGE_RESISTANCE_WINDOW` and the getter prototype
- `src/core/prefs.c` — wire up the pref, its string name, and `meta_prefs_get_edge_resistance_window()`
- `src/core/edge-resistance.c` — use the pref to gate the towards-window pixel threshold

## Testing

Built from source and installed as a .deb on Debian Trixie (Cinnamon).
With the key set to `false`, windows drag freely with no edge
resistance; back to `true`, resistance behaves as before.

## Scope

The toggle only affects resistance between **window** edges. Resistance
against **monitor and screen boundaries** is intentionally left intact
and stays active regardless of this setting. This matches the original
behavior of the old key and keeps screen-edge snapping/tiling working as
expected.

## Note

This restores the underlying setting only. Adding a checkbox back to the
Cinnamon settings GUI lives in `linuxmint/cinnamon` and would be a
separate follow-up.

### DIFF
--- a/data/org.cinnamon.muffin.gschema.xml.in
+++ b/data/org.cinnamon.muffin.gschema.xml.in
@@ -43,6 +43,17 @@
       </description>
     </key>
 
+    <key name="edge-resistance-window" type="b">
+      <default>true</default>
+      <summary>Enable edge resistance between windows</summary>
+      <description>
+        If enabled, windows resist movement when their edges approach the
+        edges of other windows. Disabling this allows windows to be dragged
+        freely past one another without any sticky edge effect. Resistance
+        against monitor and screen boundaries is not affected by this setting.
+      </description>
+    </key>
+
     <key name="tile-maximize" type="b">
       <default>false</default>
       <summary>Sets maximize as the tile action for the top edge of the screen</summary>
--- a/src/core/edge-resistance.c
+++ b/src/core/edge-resistance.c
@@ -27,6 +27,7 @@
 #include "core/display-private.h"
 #include "core/meta-workspace-manager-private.h"
 #include "core/workspace-private.h"
+#include "meta/prefs.h"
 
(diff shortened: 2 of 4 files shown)


ianjosephthompson:
+1 Very eager for this to return.

PR #830, opened 2026-06-19
clutter: Round widget coordinates and dimensions

clefebvre:
Clutter widgets look blurry when they are positioned on fractional coordinates or when their width or height are not rounded.

Whenever Cinnamon calls set_position(x, y), round x and y to avoid this.

Whenever Cinnamon calls set_size(w, h), ceil w and h. The reason we use ceilf instead of roundf here is to avoid potential clipping. For instance, if a text determines the dimension of an entry, we don't want to round down and take the risk to clip part of the font.

### DIFF
--- a/clutter/clutter/clutter-actor.c
+++ b/clutter/clutter/clutter-actor.c
@@ -10625,7 +10625,7 @@ clutter_actor_set_position (ClutterActor *self,
 
   g_return_if_fail (CLUTTER_IS_ACTOR (self));
 
-  graphene_point_init (&new_position, x, y);
+  graphene_point_init (&new_position, roundf (x), roundf (y));
 
   cur_position.x = clutter_actor_get_x (self);
   cur_position.y = clutter_actor_get_y (self);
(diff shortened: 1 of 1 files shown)


clefebvre:
Before:




After:

clefebvre:
Before:

image

After:

image

clefebvre:
The following were confirmed to always produce fractional coordinates:

- Popup menus (popupMenu.js): every menu in Cinnamon was positioned at a fractional x, because _calculatePosition centers the menu on its source actor using two successive halvings ((x2 - x1) / 2 then natWidth / 2), making a fractional result virtually guaranteed.
- Expo "Add workspace" button: y position computed as (monitorHeight - buttonHeight) / 2.
- Expo window-close drop area: x position computed as (monitorWidth - closeAreaWidth) / 2.
- Magnifier crosshairs (all four hair actors, plus the crosshair overlay on the main screen): centered using groupWidth / 2 and groupHeight / 2.
- Window app icons in expo: position computed as ICON_OFFSET / thumbnail_scale / scale (two divisions).

The following produce fractional coordinates only when global.ui_scale is set to a value where panelHeight × ui_scale is not a whole number (e.g. a 27px panel at 1.5× scale = 40.5px):

- Panel positioning (panel.js): panel height is stored as an integer setting and multiplied by ui_scale at runtime, then used directly in set_position.

mtwebster:
global.ui_scale is always a whole number, btw

PR #831, opened 2026-06-20
wayland: Fix GPU acceleration (wl_drm)

clefebvre:
This is a collection of multiple fixes to bring acceleration to Wayland and XWayland.

It's only a temporary solution for now using wl_drm implementation. We need a better implementation which supports zwp_linux_dmabuf_v1 v4/5.

Fixes XWayland acceleration, acceleration in games like Supertuxkart, playing videos in Slack... I think Vulkan already worked before that. Doesn't fix acceleration in direct GL apps such as glmark2-es2-wayland.


leigh123linux:
@clefebvre  looks like you forgot to 'git add' the new files.

clefebvre:
wl_drm implementation merged in master, PRIME and cursor commits might come back in new PRs.

PR #832, opened 2026-06-22
Gate accelerometer auto-rotation on tablet mode (X11 + Wayland)

ggiesen:
## What

Cinnamon already auto-rotates the built-in panel from the accelerometer (Muffin's `MetaOrientationManager` + `MetaMonitorManager`), gated only by `orientation-lock`. On a convertible that means it rotates in **every** posture - it never checks the tablet-mode switch, so the screen flips while the device is used as a laptop.

This gates auto-rotation on tablet mode:

- `apply_orientation()` now consults `ClutterSeat:touch-mode` (true only when a touchscreen is present **and** the device is in tablet mode or has no tablet-mode switch) plus a new `disable-rotation-in-laptop-mode` key, and honours `orientation-lock` on the touch-mode / setting-change paths.
- The **native/Wayland** backend already exposes `touch-mode` from libinput. The **X11** backend had no source for `SW_TABLET_MODE`, so `meta-seat-x11.c` now reads the evdev switch directly (by capability, with hot-plug recovery and the login-ACL race handled) and folds it into `touch-mode` - so the same gate works under X11, where Cinnamon mostly runs today.

Devices with no tablet-mode switch (slates) are unaffected - `touch-mode` is true, so they keep auto-rotating.

## Context / prior art

This is the feature from **#10697** (cinnamon) / **cinnamon-settings-daemon#359**. The earlier PRs - **cinnamon#10708** and **cinnamon-settings-daemon#361** - implemented it in the **csd orientation plugin**, which was being removed as the orientation work moved into Muffin (the reason they were closed). This does it in Muffin, where rotation now lives.

Two small **companion changes** are needed (happy to open as linked PRs):
- **cinnamon-settings-daemon**: add `disable-rotation-in-laptop-mode` (bool, default `true`) to the `…peripherals.touchscreen` schema (5 lines).
- **cinnamon**: surface it in Display → Settings. In #10697 @mtwebster suggested a 3-way combobox (*Disabled / Enabled / Tablet mode only*) instead of a second toggle — glad to do exactly that.
- X11 needs a `uaccess` udev rule so Muffin can read the switch as the session user (rule included in the companion repo below).

## Validation

Built as patched packages and dogfooded on an HP ZBook Studio x360 G5 (Mint 22.3, Cinnamon 6.6.7, Muffin 6.6.3, X11): rotation **and** tablet-mode gating handled entirely by Muffin, no external daemon. Rebased onto and applies cleanly to `master` (6.7.3). Full patch set, udev rule, and build notes: https://gitlab.com/ggiesen/cinnamon-autorotate

## Feedback welcome

This is a proposal as much as a patch - **happy to take feedback or change direction** on any of it: the combobox-vs-toggle UX, the key name/layout, splitting the X11 seat change into its own commit, or sending the Wayland-gating part upstream to GNOME Mutter as well. Just point me at what fits Cinnamon best.

### DIFF
--- a/src/backends/meta-monitor-manager-private.h
+++ b/src/backends/meta-monitor-manager-private.h
@@ -140,6 +140,12 @@ struct _MetaMonitorManager
   gulong experimental_features_changed_handler_id;
 
   MetaMonitorSwitchConfigType current_switch_config;
+
+  /* org.cinnamon.settings-daemon.peripherals.touchscreen, used to gate
+   * accelerometer auto-rotation on tablet mode. NULL when the
+   * disable-rotation-in-laptop-mode key is not present (older
+   * cinnamon-settings-daemon), in which case rotation is never gated. */
+  GSettings *orientation_settings;
 };
 
 /**
--- a/src/backends/meta-monitor-manager.c
+++ b/src/backends/meta-monitor-manager.c
@@ -67,6 +67,11 @@
 // is 0.00000011920928955078 which is less than the difference of those two scales.
 #define CINNAMON_SCALE_EPSILON 0.000001f
 
+/* Accelerometer auto-rotation: gate it on tablet mode via this setting (see
+ * apply_orientation). The schema is owned by cinnamon-settings-daemon. */
+#define ORIENTATION_SCHEMA "org.cinnamon.settings-daemon.peripherals.touchscreen"
+#define DISABLE_LAPTOP_ROTATION_KEY "disable-rotation-in-laptop-mode"
+
 enum
 {
   PROP_0,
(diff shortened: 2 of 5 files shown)


ggiesen:
Companion PRs now open: schema key in cinnamon-settings-daemon#463, and the Display *Automatic screen rotation* combobox in cinnamon#13826 (the UI @mtwebster suggested in linuxmint/cinnamon#10697). The three go together.

ggiesen:
**Field-testing note** (no code changes - recording behaviour on a second platform).

I've tested this on two HP convertibles, both Linux Mint / X11:

- **ZBook Studio x360 G5** - has a reliable `SW_TABLET_MODE` fold switch; *Tablet mode only* works as intended (holds landscape as a laptop, follows the accelerometer when folded into tablet).
- **Pavilion x360 14-dh1008ca** - no reliable fold switch. This machine's firmware uses the *screen-rotation accelerometer* (the same sensor that drives auto-rotate) as its tablet-mode indicator rather than the hinge: intel-hid's `SW_TABLET_MODE` asserts on screen **orientation** (a tilt toward portrait), not on fold - the kernel `dual_accel_detect()` gap for ISH-hosted accelerometers - and its other (HP-WMI) switch never asserts. The firmware leans all the way into this: it locks out the internal keyboard and touchpad whenever the screen is rotated to portrait, *even in laptop posture*. So the machine's own notion of "tablet" genuinely *is* orientation, which is why erring toward rotating fits this hardware rather than fighting it.

Caveat worth documenting: **"Tablet mode only" gates correctly only where the hardware exposes a real fold switch.** On a machine without one it cannot distinguish laptop from tablet, so the right thing is to err toward rotating - i.e. use **"Enabled"**, which bypasses the gate entirely (the same escape-hatch role as KWin/KScreen's tri-state *Always*). On no-reliable-sensor hardware *Tablet mode only* effectively degrades to *Enabled*/Always behaviour, which is the desired failure direction. I specifically do **not** recommend `intel_hid.enable_sw_tablet_mode=0` there: it removes the only switch that ever asserts and would stop rotation altogether.

For completeness: I considered and rejected compositor-side heuristics to auto-detect bad switches - "rotate until the switch toggles" would reintroduce laptop-mode rotation at every cold boot, and persisting a "proven" flag adds state yet still wouldn't fix the orientation-coupled case (that switch does toggle). The stateless read here mirrors the native/libinput backend, so no change is warranted.

PR #839, opened 2026-07-09
[Wayland] Avoid using buffers with unfinished client work

Sunderland93:
Backported and adapted from Mutter.

See: https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1880

In addition to what is stated in the original MR description in Mutter, this also sets the stage for the implementation of the `fifo-v1` and `linux-drm-syncobj-v1` protocols


Sunderland93:
@mtwebster please review it

PR #840, opened 2026-07-12
[Wayland] Switch to stable linux-dmabuf

Sunderland93:


Fantu:
Hi, I saw that old [NVIDIA EGLStream controller](https://wayland.app/protocols/nvidia-eglstream-controller) protocol is still used only in muffin.
Now recent driver already prefer and use linux-dmabuf, so the old nvidia protocol can be removed, or I'm wrong?

Sunderland93:
Hi. `EGLStream` controller is not replaced by `linux-dmabuf`, it's related to buffer management in driver, and recently Muffin is switching to use `GBM` by default. But Muffin doesn't contains many many Nvidia-related fixes that Mutter has (one of them https://github.com/linuxmint/muffin/pull/839 which is needed to properly implement `drm-syncobj` protocol which in turn is very important for Nvidia driver). So I prefer to leave EGLStream controller support in place, until Muffin get all necessary Nvidia pieces.

Fantu:
I think didn't check whether the Wayland (and protocol) versions in meson and d/control need to be updated. d/control doesn't seem to be updated.
Anyway I don't have time now to check all the Wayland protocols and functions currently used in Muffin to see what minimum versions need to be set.

PR #842, opened 2026-07-14
[Wayland] Implement ext-idle-notify-v1 protocol

Sunderland93:
This PR implements the `ext-idle-notify-v1` Wayland protocol, allowing desktop components to receive user idle/active notifications, which is useful for applications that track user activity. It leverages Muffin's existing `MetaIdleMonitor` backend, extending it to support bypassing inhibitors.


Sunderland93:
Tested with `swayidle timeout 10 nemo` (`nemo` should start in 10 seconds)

PR #793, opened 2026-02-08
Cleanup and refactor resource scale API

Sunderland93:
See individual commit messages

This is the preparation for https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1285 which is needed to drive each output on it's own refresh rate.

It's also improves performance and quality on HiDPI displays.

Cinnamon part: https://github.com/linuxmint/cinnamon/pull/13542
Cinnamon-Spices-Applets: https://github.com/linuxmint/cinnamon-spices-applets/pull/8293

PR #843, opened 2026-07-17
[Wayland] Add support for presentation-time protocol

Sunderland93:
This PR add support for `presentation-time` protocol, which is needed to get proper timestamp feedback from application for improved frame pacing. At this time Muffin does not support separate frame clock's per CRTC, so this implementation is tied to the single master clock (my initial attempt to prepare support for separate frame clock's is here https://github.com/linuxmint/muffin/pull/793)

Tested with `weston-presentation-shm` and MPV.


Sunderland93:
Since #793 is merged - there is no need in single frame-clock based approach. We can backport complete implementation of presentation-time from GNOME

PR #854, opened 2026-08-18
Fix desktop zoom scrolling the window under the pointer, and Super opening the menu after zooming

### LINKED ISSUE #695: Desktop Zoom Mouse Wheel scrolling also scrolls XInput2 application window contents
### Distribution

Manjaro 24.0.2

### Package version

6.0.1

### Graphics hardware in use

AMD Ryzen 7 7840HS w/ Radeon 780M Graphics

### Frequency

Always

### Bug description

When the mouse pointer is over an application that uses XInput2 for scroll event handling, **Desktop Zoom** is enabled, and a **Mouse wheel modifier** is set for it, zooming in also scrolls the window contents up, and zooming out scrolls the window contents down.

Non-XInput2 applications are not affected, but this is a shrinking list as more applications adopt XInput2 for pixel-perfect scrolling, gestures, etc.

This issue has been reported a handful of times on the [Cinnamon issue tracker](https://github.com/linuxmint/cinnamon/issues/), but I believe it is a bug in Muffin specifically.

Related Cinnamon issue tickets:

* [Mouse wheel zoom also scrolls content #5163](https://github.com/linuxmint/cinnamon/issues/5163) (closed "works for me", I think because XInput2 involvement hadn't been identified)
* [[Request] Desktop zoom improvement #10084](https://github.com/linuxmint/cinnamon/issues/10084)
* [Desktop zoom scrolls underlying content #10148 ](https://github.com/linuxmint/cinnamon/issues/10148)
* [[BUG] Scrolling up/down while zooming in/out #11728](https://github.com/linuxmint/cinnamon/issues/11728)

### Steps to reproduce

1. Open `cinnamon-settings` **Accessibility → Visual**
2. Enable **Desktop Zoom → Enabled**
3. Set `<Alt>` as the **Desktop Zoom → Mouse wheel modifier** (other modifiers have the same problem, but scroll events may cause different behaviors depending on the modifier used)
4. Position the mouse cursor over a window for an XInput2-aware application with scrollable contents. (e.g. Chrome, Chromium, any Electron app, `cinnamon-settings`, Firefox if `MOZ_USER_XINPUT2=1` is set, Emacs 29.1 if compiled with `--use-xinput2`)
5. Hold `<Alt>`
6. Send scroll events (e.g. spin your mouse's scroll wheel, two-finger-drag your touch pad)

### Expected behavior

The Desktop Zoom magnification level changes, but the contents of the window under the mouse pointer does not scroll.

**Actual behavior:**
The application under the mouse cursor reacts to the scrolling events as well. The modifier being held may affect how the application responds. For example, Emacs will scroll window contents when `<Alt>` is the modifier, but not `<Super>`, and Firefox will scroll the window contents with `<Super>` but cycle through the browser history with `<Alt>`.

(Added bonus bug: if `<Super>` is the selected modifier, and is also the menu activation modifier, the Cinnamon menu will pop up when you release `<Super>`, even if you've used the mouse wheel to modify the Desktop Zoom magnification)

### Additional information

My evidence that this is related to XInput2 support:

* Firefox can be told whether or not to use XInput2 by setting the environment variable `MOZ_USE_XINPUT2=1`. This bug only affects Firefox for me when using XInput2.
* I made two builds of Emacs 29.1, with the only difference being whether `./configure` was called with `--with-xinput2` or `--without-xinput2`. This bug effects the `--with-xinput2` build, but not the `--without-xinput2` build.
* `xev`, which reports Core Pointer events (buttons 4 and 5 for scrolling), doesn't report any scroll events when the zoom modifier is held down and I scroll to change the zoom level. `xinput test-xi2 2` however, continues to receive valuator 3 `XI_Motion` events regardless of whether I'm holding down the zoom modifier.

I don't understand completely how to resolve this, but I have some guesses about why it's happening. Prior to XInput2, scroll up and down events were reported as mouse buttons 4 and 5 `ButtonPress`/`ButtonRelease` event pairs. These events are properly grabbed by `src/core/keybindings.c:meta_display_window_buttons()`, which sets up a passive grab on Mod+Button4 and Mod+Button5.

XInput2 applications, however, get scroll events as `XI_Motion` events on a valuator associated with a vertical type `XIScrollClass` (valuator 3 for the default Virtual Core Pointer, at least on my machine). These don't seem to be grabbed successfully. `meta_change_button_grab()` does include `XI_Motion` in the mask of events it passes to `XIGrabButton()`, but I don't think `XIGrabButton()` allows setting up passive grabs on `XI_Motion` events that way.

AlexB7:
Fixes #695
Addresses linuxmint/cinnamon#12587 (and the long lineage behind it: linuxmint/cinnamon#5163, #10084, #10148, #11728)

## The bugs

With **Accessibility → Zoom** enabled and a mouse-wheel modifier set, two things go wrong on X11:

1. **Scroll passthrough**: zooming also scrolls the window under the pointer — vertically with Super, horizontally with Shift, app-zoom with Ctrl. As diagnosed in #695, this only affects XInput2-aware clients (GTK3 apps, Chromium/Electron, Firefox with `MOZ_USE_XINPUT2=1`), which is by now nearly everything.
2. **Bonus bug from #695**: with Super as the modifier, releasing Super after zooming pops up the Cinnamon menu.

Note:

1. These issues do not affect Linux Mint 22.3 Wayland except for the super key bringing up the menu.  The first commit solves this issue for both Wayland and X11.
2. With a stable Wayland coming in the next release of Mint, one could argue that this very old bug doesn't really need to be fixed but end of support for 22.x isn't until April 2029.

## Root causes

**Passthrough**: muffin's zoom capture uses passive `XIGrabButton` grabs on buttons 4/5. Those only ever intercept the *emulated legacy* scroll button events. With XI 2.1 smooth scrolling, the X server delivers the smooth-scroll `XI_Motion` valuator event to the client under the pointer *before* the emulated button event activates the passive grab — motion events cannot trigger a passive button grab. So muffin gets the button event (zoom works) **and** the client gets the motion event (window scrolls). This was confirmed with `xev` vs `xinput test-xi2`, and with Firefox/Emacs XInput2 on/off builds (see #695).

**Menu popup**: scroll events consumed by the zoom branch in `events.c` bypass `meta_keybindings_process_event`, so the modifier-only ("overlay key") state is never cancelled; Super-press → zoom → Super-release looks like a bare Super tap.

## The fix

**Commit 1** cancels the pending modifier-only state when a zoom scroll is consumed (and on scroll events generally, matching the existing button/touch behavior). Fixes the menu popup on both X11 and Wayland.

**Commit 2** takes an **active grab on the virtual core pointer for the duration of a zoom scroll burst**. The first consumed zoom scroll engages the grab (via the existing `meta_backend_grab_device()`); each further scroll refreshes a 1 s idle timer. The grab is released on timeout, on a button press, or as soon as the zoom modifier is released (tracked through `XkbStateNotify`, which muffin already receives — no key grabs involved). While the grab is held, smooth-scroll motion events route to muffin and cannot reach the client.

The grab is intentionally *not* held for the whole time the modifier is down. Modifier+click interactions (Shift+click selection, Ctrl+click) must keep working, and re-injecting a swallowed click via XTest is not possible: the physical press has already registered in the master device's button state, so the synthetic press is discarded by the server as a duplicate (verified with `xev`: no `ButtonPress` delivered, `EnterNotify state 0x111`).

Wayland is unaffected by commit 2 (guarded, and not needed — muffin consumes the events before delivery there).

## Known limitations

- The **first scroll event of each burst** can still leak to an XI2 client (it has been processed by the time the grab can engage). Subsequent events in the burst are fully captured.  Using the Super key may be the best choice as it won't interfere with side scrolling, browser zooming etc.
- The engage can fail transiently (`AlreadyGrabbed`) if another client holds a pointer grab (open menu, button held); it retries on the next scroll event and zoom itself keeps working through the passive grabs.
- A click landing within the 1 s idle window after the last zoom scroll is consumed by the grab teardown.

## Testing this PR on the stable 6.6.x series (Mint 22.3)

This PR targets master, but the patches apply cleanly to the current stable release, so anyone affected can try them on a **test installation of** Linux Mint 22.x / muffin 6.6.x:

**Pre-rebased branch** (easiest): a backport onto the `6.6.3` tag is at [`AlexB7/muffin` branch `zoom-fix-6.6.3`](https://github.com/AlexB7/muffin/tree/zoom-fix-6.6.3):

```bash
git clone --branch zoom-fix-6.6.3 https://github.com/AlexB7/muffin
cd muffin
sudo apt build-dep muffin        # needs a deb-src line enabled
dpkg-buildpackage -us -uc -b
sudo apt install ../libmuffin0_*.deb ../muffin_*.deb ../muffin-common_*.deb ../gir1.2-meta-muffin-0.0_*.deb
```

**Or apply this PR's commits yourself**:

```bash
git clone https://github.com/linuxmint/muffin && cd muffin
git checkout -b zoom-test 6.6.3
curl -L https://github.com/linuxmint/muffin/pull/854.patch | git am -3
```

Either way, install the full package set (the fix lives in `libmuffin0`) and log out/in afterwards. To verify: enable **Accessibility → Zoom** with a mouse wheel modifier, hold the modifier and scroll over Chromium or any GTK3 app — the window content should no longer scroll while zooming (apart from at most the first tick of a burst), and with Super as the modifier the menu should no longer open when you release the key.

**Reverting**: the locally built version (`6.6.3`) sorts below the distro's (`6.6.3+<codename>`), so going back to stock is just

```bash
sudo apt install --reinstall --allow-downgrades muffin libmuffin0 muffin-common gir1.2-meta-muffin-0.0
```

and a log out/in. For the same reason, Update Manager will offer to replace the patched build with the stock package at the next update — hold the packages (`sudo apt-mark hold muffin libmuffin0 muffin-common gir1.2-meta-muffin-0.0`) if you want to keep testing through updates, and `unhold` when done.

## Testing done

On Linux Mint 22.3, muffin 6.6.3 + these patches, X11:

- Super/Shift + wheel over Chromium, GNOME Terminal, and `MOZ_USE_XINPUT2=1` Firefox: zoom works, window content no longer scrolls (previously reproduced reliably in all three).
- Super release after zooming no longer opens the menu (also verified over VNC, where only the legacy-button path exists).
- Shift as modifier: Shift+click and Shift+drag text selection in xed work with zoom enabled; Shift horizontal scroll correctly suppressed while zooming.
- Plain Super tap still opens the menu; Alt-Tab, window drag, unmodified scrolling unchanged.
- Wayland session: no regressions; passthrough does not occur there by design.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/backends/x11/meta-backend-x11.c
+++ b/src/backends/x11/meta-backend-x11.c
@@ -53,6 +53,7 @@
 #include "clutter/x11/clutter-x11.h"
 #include "compositor/compositor-private.h"
 #include "core/display-private.h"
+#include "core/keybindings-private.h"
 #include "meta/meta-cursor-tracker.h"
 #include "meta/util.h"
 
@@ -425,6 +426,12 @@ handle_host_xevent (MetaBackend *backend,
                     meta_backend_notify_keymap_layout_group_changed (backend,
                                                                      layout_group);
                 }
+              if (display &&
+                  (xkb_ev->state.changed & (XkbModifierStateMask |
+                                            XkbModifierBaseMask)))
+                meta_display_process_zoom_modifier_state (display,
+                                                          xkb_ev->state.mods,
+                                                          xkb_ev->state.time);
               break;
             default:
               break;
@@ -579,6 +586,14 @@ meta_backend_x11_post_init (MetaBackend *backend)
                                     &priv->xkb_error_base))
     meta_fatal ("X server doesn't have the XKB extension, version %d.%d or newer\n",
                 XKB_X11_MIN_MAJOR_XKB_VERSION, XKB_X11_MIN_MINOR_XKB_VERSION);
+
+  /* Make sure modifier state changes are among the selected
+   * XkbStateNotify details; the zoom modifier pointer grab depends on
+   * them (see meta_display_process_zoom_modifier_state).  Only the
+   * listed detail bits are affected, existing selections remain. */
+  XkbSelectEventDetails (priv->xdisplay, XkbUseCoreKbd, XkbStateNotify,
+                         XkbModifierStateMask | XkbModifierBaseMask,
+                         XkbModifierStateMask | XkbModifierBaseMask);
 
   META_BACKEND_CLASS (meta_backend_x11_parent_class)->post_init (backend);
 
--- a/src/core/events.c
+++ b/src/core/events.c
@@ -291,6 +291,10 @@ meta_display_handle_event (MetaDisplay        *display,
             meta_display_a11y_zoom (display, FALSE);
           }
 
+        /* Don't let the zoom modifier's release trigger a
+         * modifier-only keybinding (e.g. Super opening the menu) */
+        meta_keybindings_cancel_modifier_only (display);
+
         bypass_wayland = bypass_clutter = TRUE;
         goto out;
       }
--- a/src/core/events.c
+++ b/src/core/events.c
@@ -295,10 +295,19 @@ meta_display_handle_event (MetaDisplay        *display,
          * modifier-only keybinding (e.g. Super opening the menu) */
         meta_keybindings_cancel_modifier_only (display);
 
+        /* On X11, hold the pointer for the duration of the scroll burst
+         * so smooth-scroll events can't leak to the client under the
+         * pointer. */
+        meta_display_zoom_scroll_grab_notify (display,
+                                              clutter_event_get_time (event));
+
         bypass_wayland = bypass_clutter = TRUE;
         goto out;
       }
     }
+
+  if (event->type == CLUTTER_BUTTON_PRESS)
+    meta_display_zoom_grab_break (display, clutter_event_get_time (event));
 
   if (event->type != CLUTTER_DEVICE_ADDED &&
       event->type != CLUTTER_DEVICE_REMOVED)
--- a/src/core/keybindings-private.h
+++ b/src/core/keybindings-private.h
@@ -144,6 +144,7 @@ int      meta_keybindings_get_mouse_zoom_modifiers (MetaDisplay *display);
 ClutterModifierType meta_display_get_window_grab_modifiers (MetaDisplay *display);
 
 uint     meta_keybindings_get_ignored_modifier_mask (MetaDisplay *display);
+void     meta_keybindings_cancel_modifier_only (MetaDisplay *display);
 
 gboolean meta_prefs_add_keybinding          (const char           *name,
                                              GSettings            *settings,
--- a/src/core/keybindings-private.h
+++ b/src/core/keybindings-private.h
@@ -126,6 +126,10 @@ typedef struct
   /* Alt+click button grabs */
   ClutterModifierType window_grab_modifiers;
   ClutterModifierType mouse_zoom_modifiers;
+
+  /* Active pointer grab held during zoom scroll bursts (X11) */
+  gboolean zoom_pointer_grabbed;
+  guint zoom_grab_timeout_id;
 } MetaKeyBindingManager;
 
 void     meta_display_init_keys             (MetaDisplay *display);
@@ -145,6 +149,14 @@ ClutterModifierType meta_display_get_window_grab_modifiers (MetaDisplay *display
 
 uint     meta_keybindings_get_ignored_modifier_mask (MetaDisplay *display);
 void     meta_keybindings_cancel_modifier_only (MetaDisplay *display);
+
+void     meta_display_process_zoom_modifier_state (MetaDisplay *display,
+                                                   unsigned int mods,
+                                                   guint32      timestamp);
+void     meta_display_zoom_scroll_grab_notify (MetaDisplay *display,
+                                               guint32      timestamp);
+void     meta_display_zoom_grab_break (MetaDisplay *display,
+                                       guint32      timestamp);
 
 gboolean meta_prefs_add_keybinding          (const char           *name,
                                              GSettings            *settings,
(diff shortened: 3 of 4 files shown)


malsyned:
@AlexB7 thank you so much for working on this, it's been bothering me for so long! Manjaro Cinnamon is currently on 6.6.3, so that's the version I applied your two patches to.

The bug is hugely improved by this change, but not quite fixed, unfortunately. It looks like some motion events are still making it through to the application at the beginning of the zoom operation before the capture kicks in. This is a minor nuisance with a lot of programs, but it's particularly annoying in Firefox when using Alt as the zoom modifier, because Firefox uses Alt+scroll to navigate back and forward in the history.

To reproduce:
1. Make sure modifier key is released
2. Hover over XInput2-aware window
3. Press modifier
4. two-finger drag up or down
5. observe underlying window scroll one notch, then stop scrolling as you keep dragging and the zoom starts

A possibly useful detail is that, at step 4 above, if I drag very very slowly and carefully, I can scroll the underlying window up and down in a small range of values (e.g. about ±3 lines of text in gnome-terminal) indefinitely without the zoom engaging and starting to capture the motion events. As soon as I pass some threshold, the zoom starts and the window underneath stops scrolling.

If I release the modifier before scrolling more than that threshold amount, the modifier-only state isn't canceled either. For example, if my modifier is set to Super, the system menu pops up.

Thanks again for giving this issue some love, let me know if you have updated revs that you want me to test for you.

malsyned:
I did a little screen recording showing the ability to scroll around within the zoom trigger threshold.

desktop-zoom-bug

It also shows something funky with the cursor, where it seems to be fighting with my use of Super as the window drag modifier. When I turn the zoom functionality off and hold Super, the cursor becomes the four-pointed cross when I start two-finger dragging, and goes back to being a pointer arrow when I take my fingers off the trackpad. With Zoom on, it looks like the window drag mode isn't getting the drag release event.

malsyned:
Oh whoops, I just saw your "known limitations" section where you document that you already know about this issue. Sorry for the noise!

